Issue #6, opened 2026-04-18
Windows support


vedran-:
Any plans to support Windows?

mlhher:
Since I do not use Windows I cannot add the support myself.
The main issue is the bash tool requiring a heavy rework.

If possible try and use WSL, otherwise this needs external attention. Will leave it open for now.

SRes:
It's fairly trivial by using `git bash`, pretty much plug-and-play solution. I've managed to run it on Windows in 30 minutes.
However, prompts will need to be adapted, Qwen 3.6 immediately tries to use Windows commands seeing Windows paths, but being in a bash emulator, it fails. Some tests break because of Linux assumptions, minor issues.

aa956:
There are actually issues with Windows sources even if late itself runs on Linux.

@mlhher - should it be filed as separate bug with `target_edit` tool? If source file contains Windows line ends, target_edit fails.

Example (edited filenames and paths):

```text
Thinking Process:
    ┃ The user wants me to fix a bug in SourceFile.c at line 344 where green_byte is checked twice instead of blue_byte. I need to read the file first to confirm the exact line and then
    ┃ make the fix using target_edit.
    ┃
  ◆ Reading file ./src/SourceFile/SourceFile.c
  Thinking Process:
    ┃ I found the bug at line 344 where it incorrectly assigns green_byte instead of blue_byte. Now I need to use target_edit to fix this specific line.
    ┃
 │    I can see the bug at line 344. The  if  condition correctly checks  blue_byte
 │    but then incorrectly assigns  green_byte  to  max_byte . Let me fix this:
 │
  ◆ Editing file SourceFile.c
  Thinking Process:
    ┃ The search block failed due to whitespace mismatch - let me adjust the indentation to match the actual file content.
    ┃
 │    Let me check the exact indentation:
 │
  ◆ Editing file SourceFile.c
  Thinking Process:
    ┃ The previous search block didn't match the file exactly, likely due to whitespace differences. Let me read more context around the target lines to capture the exact formatting.
    ┃
  ◆ Reading file ./src/SourceFile/SourceFile.c
  Thinking Process:
    ┃ I need to adjust my search block to match the exact indentation shown in the file (8 spaces) and include more context to ensure uniqueness.
    ┃
 ◆ Editing file SourceFile.c
  Thinking Process:
    ┃ The search block keeps failing due to whitespace mismatch, so I need to include more surrounding context to ensure an exact match.
    ┃
[...]

PR #19, opened 2026-04-21
feat: Windows support and cross-platform compilation

giveen:
## Summary
Adds full Windows support to `late`, enabling cross-platform compilation and
execution on both Windows and Linux/macOS.

### Platform-aware shell execution
- Added `shell_command_windows.go` — uses `cmd.exe /C` on Windows
- Added `shell_command_unix.go` — resolves `bash`/`sh` at runtime with lazy
  caching, falling back gracefully if `bash` is not found
- Build tags (`//go:build windows` / `//go:build !windows`) ensure the correct
  implementation is compiled per platform

### Cross-platform path utilities
- Added `path_utils.go` with `LateConfigDir()` and `LateSessionDir()`:
  - Windows: stores config and sessions under `%AppData%\late` via `UserConfigDir()`
  - Linux/macOS: sessions stored under `~/.local/share/late/sessions` (XDG convention)
- Added `path_utils_test.go` with platform-conditional assertions

### General fixes & polish
- Updated `config.go`, `executor.go`, `mcp/config.go`, `session/models.go`,
  and TUI interactions to use the new path helpers
- Renamed shell helper variables for clarity; aligned docs examples
- Added tests for the shell command abstraction layer

## Testing
- All existing tests pass on both Windows and Linux
- New tests cover: path resolution per OS, shell command construction,
  Unix shell fallback behavior
  
  
image

### DIFF
--- a/internal/tool/implementations.go
+++ b/internal/tool/implementations.go
@@ -334,7 +334,7 @@ func (t *ShellTool) WrapError(ctx context.Context, err error) error {
 
 	var errorMsg string
 	if strings.Contains(strings.ToLower(orchestratorID), "coder") {
-		errorMsg = fmt.Sprintf("Do not use a %s command like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", shellDisplayName(), err.Error())
+		errorMsg = fmt.Sprintf("Do not use %s commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", shellDisplayName(), err.Error())
 	} else {
 		errorMsg = fmt.Sprintf("You are an architect/planner agent. You cannot write files. To modify files, you must spawn a coder subagent using `spawn_subagent` tool. %s", err.Error())
 	}
--- a/internal/tool/implementations.go
+++ b/internal/tool/implementations.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
@@ -432,6 +433,15 @@ func (t ShellTool) Execute(ctx context.Context, args json.RawMessage) (string, e
 		return "", t.WrapError(ctx, err)
 	}
 
+	// Enforce approval in the execution path so shell commands fail closed
+	// even if middleware wiring is missing.
+	if t.RequiresConfirmation(args) {
+		approved, ok := ctx.Value(common.ToolApprovalKey).(bool)
+		if !ok || !approved {
+			return "", fmt.Errorf("shell command requires explicit approval before execution")
+		}
+	}
+
 	// Validate and set working directory
 	if params.Cwd != "" {
 		if !IsSafePath(params.Cwd) {
@@ -475,6 +485,11 @@ func (t ShellTool) RequiresConfirmation(args json.RawMessage) bool {
 		return true // Default to requiring confirmation if we can't parse
 	}
 
+	// Conservative Windows policy: always require confirmation.
+	if runtime.GOOS == "windows" {
+		return true
+	}
+
 	// If the command contains any shell metacharacters that could embed
 	// sub-commands or disguise intent, always require confirmation.
 	// This is the primary defense: we don't try to parse complex shell
@@ -500,11 +515,19 @@ func (t ShellTool) CallString(args json.RawMessage) string {
 		Cwd     string `json:"cwd"`
 	}
 	if err := json.Unmarshal(args, &params); err != nil {
+		if runtime.GOOS == "windows" {
+			return "Executing in PowerShell: (invalid args)"
+		}
 		return "Executing: (invalid args)"
 	}
 
 	// Build the display string
-	result := fmt.Sprintf("Executing: %s", params.Command)
+	var result string
+	if runtime.GOOS == "windows" {
+		result = fmt.Sprintf("Executing in PowerShell: %s", params.Command)
+	} else {
+		result = fmt.Sprintf("Executing: %s", params.Command)
+	}
 	if params.Cwd != "" {
 		result += " in dir: " + params.Cwd
 	}
--- a/internal/tool/implementations.go
+++ b/internal/tool/implementations.go
@@ -334,7 +334,7 @@ func (t *ShellTool) WrapError(ctx context.Context, err error) error {
 
 	var errorMsg string
 	if strings.Contains(strings.ToLower(orchestratorID), "coder") {
-		errorMsg = fmt.Sprintf("Do not use %s commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", shellDisplayName(), err.Error())
+		errorMsg = fmt.Sprintf("Do not use a %s command like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", shellDisplayName(), err.Error())
 	} else {
 		errorMsg = fmt.Sprintf("You are an architect/planner agent. You cannot write files. To modify files, you must spawn a coder subagent using `spawn_subagent` tool. %s", err.Error())
 	}
--- a/internal/tool/implementations.go
+++ b/internal/tool/implementations.go
@@ -334,7 +334,7 @@ func (t *ShellTool) WrapError(ctx context.Context, err error) error {
 
 	var errorMsg string
 	if strings.Contains(strings.ToLower(orchestratorID), "coder") {
-		errorMsg = fmt.Sprintf("Do not use host-native shell commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", err.Error())
+		errorMsg = fmt.Sprintf("Do not use %s commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", shellDisplayName(), err.Error())
 	} else {
 		errorMsg = fmt.Sprintf("You are an architect/planner agent. You cannot write files. To modify files, you must spawn a coder subagent using `spawn_subagent` tool. %s", err.Error())
 	}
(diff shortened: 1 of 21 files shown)


giveen:
I'll work on those issues tomorrow

mlhher:
Great work!

While skimming over the code I noticed one thing (which is specifically why the issue came up initially). The bash tool puts in significant work to decide whether a command can be auto executed or if it requires approval. Depending on your implementation it will prompt every time or never. Never prompting is unacceptable. Prompting every time is sub-optimal (but better than no Windows support and never prompting). Since I do not know how this would work on Windows I cannot help there. Instead of having to re-implement the entire logic maybe we can use Bash for Windows if it exists instead of cmd?

Further since I do not have any Windows machine it would be good if you could also add a e.g. `implementations_cmd_test.go` so we could run the Windows specific tests in a runner env.

Another point: I honestly never expected someone to contribute this quickly (thanks!) so I did not have the chance yet to setup the necessary PR template. Because this PR was opened before I added the necessary template, you need to manually confirm your agreement before this can be potentially merged in.

Thank you for your work! I am certain a lot of people would enjoy this a lot.

giveen:
I have read and agree to the terms of the Contributer License Agreement.

mlhher:
Please also check out #6 
Potentially the cmd route is actually preferable.

giveen:
@mlhher 
Please check the latest commit


```
feat(windows): switch shell backend to PowerShell, harden approval gating, add Windows test suite
- Replace Windows shell backend (shell_command_windows.go) from cmd.exe /C to
  PowerShell using -EncodedCommand (UTF-16LE Base64). Tries pwsh.exe (PS 7+)
  first, falls back to powershell.exe (Windows PS 5.1). UTF-16LE Base64 encoding
  ensures correct handling of quotes, pipes, spaces and special characters
  without manual escaping.

- Add ToolApprovalKey context marker (common/interfaces.go) and enforce fail-closed
  execution in ShellTool.Execute: every shell call requires an explicit approval
  marker in the context regardless of middleware wiring.

- Windows always requires confirmation: ShellTool.RequiresConfirmation returns
  true unconditionally on Windows (runtime.GOOS == windows).

- Block unsupervised-mode bypass on Windows in TUI middleware (interactions.go):
  Windows shell calls are always routed to the confirmation prompt even in
  unsupervised mode. Approval marker is propagated into context after the user
  confirms so ShellTool.Execute sees it.

- Update TUI confirmation wording (view.go): Windows prompt now reads
  'The agent wants to execute a PowerShell command.' instead of 'bash'.

- Update streaming activity line (implementations.go CallString): on Windows
  the running banner reads 'Executing in PowerShell: <command>'.

- Inject Windows platform note into system prompt (cmd/late/main.go): agent is
  told it is running in PowerShell and should prefer PowerShell-native cmdlets
  (Get-ChildItem, Get-Content, Remove-Item, etc.) over Unix aliases or bash.

- Add new Windows-only test file (implementations_cmd_test.go, //go:build windows):
  TestPSShellCommand_UsesPowerShell         - backend is pwsh.exe / powershell.exe
  TestPSShellCommand_HasRequiredFlags       - -NoProfile -NonInteractive -EncodedCommand present
  TestPSShellCommand_EncodedCommandDecodesCorrectly - round-trip for paths with spaces and quotes
  TestPSShellTool_WindowsAlwaysRequiresConfirmation - confirmation required for every command
  TestPSShellTool_ExecuteFailsWithoutApproval       - fail-closed: execution blocked without marker
  TestPSShellTool_ExecuteSucceedsWithApproval       - approval marker unlocks execution
  TestPSShellTool_CallString                        - streaming line uses PowerShell prefix
  Also verified that sub-agents spawned via TUI spawn middleware inherit the same
  confirmation gates and approval marker propagation.

- Clean up shell_command_test.go: restricted to //go:build !windows; removed
  stale cmd.exe Windows branch and unused imports. Windows assertions now live
  exclusively in implementations_cmd_test.go.

- Fix pre-existing Windows path escaping in implementations_test.go and
  targetEdit_test.go: use filepath.ToSlash when embedding temp paths in JSON
  strings; skip Unix-centric echo/seq/pwd tests on Windows.

- Add executor fail-closed test (executor_test.go):
  TestExecuteToolCalls_NoMiddlewareFailsClosed confirms shell execution is
  rejected when no middleware propagates the approval marker.

- Add TUI interactions test (interactions_test.go):
  TestTUIConfirmMiddleware_ConfirmedExecutionMarksApproval confirms approval
  marker is set in context after user presses y.
  ```

**Comment by giveen** [comment]
https://github.com/user-attachments/assets/06ce6804-881e-4839-a90d-174d50a2f805

**Comment by mlhher** [comment]
@giveen Might want to remove the video with your path (just referencing your previous screenshot). Great work though! Highly appreciated. I will go over it so please allow for a little time. Thank you a lot!

**Comment by giveen** [comment]
> @giveen Might want to remove the video with your path (just referencing your previous screenshot). Great work though! Highly appreciated. I will go over it so please allow for a little time. Thank you a lot!

Yeah, wish there was a way to edit that, lol, i'll remove it now, just wanted you to see its executing and sub-agents are working

**Comment by giveen** [comment]
@mlhher after this is merged, if you do so, i will be working on a github action to do a windows compile into a MSI, so that it installs and then becomes path-aware like the LInux binary verison, so I dont have to tell `late` "look at this directory"

File: internal/tool/implementations.go [hunk 1]
```diff
@@ -289,83 +290,84 @@ func isCdCommand(command string) (bool, error) {
 	if idx := strings.Index(cleanCmd, "#"); idx != -1 {
```
mlhher, line 290: To verify: Do Linux and Windows have the exact same behavior here?

internal/tool/implementations.go
```diff
@@ -289,83 +290,84 @@ func isCdCommand(command string) (bool, error) {
...
 		return nil
 	}
 
 	orchestratorID := common.GetOrchestratorID(ctx)
-	
+
 	var errorMsg string
 	if strings.Contains(strings.ToLower(orchestratorID), "coder") {
-		errorMsg = fmt.Sprintf("Do not use bash commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", err.Error())
+		errorMsg = fmt.Sprintf("Do not use host-native shell commands like `cat > file` or `echo > file` to write files. Use the native `write_file` or `target_edit` tools instead. %s", err.Error())
```
mlhher: Is this codepath ever called in the Windows implementation? If not it should explicitly stay with "bash" for now.

internal/tool/implementations.go
```diff
@@ -400,24 +402,24 @@ func getAllBaseCommands(command string) []string {
...
-type BashTool struct{}
+// ShellTool executes host-native shell commands with security restrictions.
+type ShellTool struct{}
 
-func (t BashTool) Name() string { return "bash" }
-func (t BashTool) Description() string {
-	return "Execute a bash command."
+func (t ShellTool) Name() string { return "bash" }
+func (t ShellTool) Description() string {
+	return "Execute host-native shell commands."
```
mlhher: This should be adjusted to inject the explicit shells name for the model. I acknowledge it requires some changes to the tool but I think preserving the models reasoning capabilities is the top priority. Even if the gain is just minimal these factors do compound massively on the long run specifically for smaller models.

"Execute a %s command." where %s = PowerShell/bash.

mlhher:
Looks good so far. I assume the msi is these install wizards I remember right?
How is this built is it like a Microsoft provided workflow? What is the licensing there can you please provide a link to the docs or similar?

internal/tool/implementations.go
mlhher: If you do the injection of the name (bash/pwsh) preferably this should also be adjusted to use the correct explicit shell name.

internal/tool/implementations.go
mlhher: Also please use singular. If we explicitly instruct it in plural it presumably is more likely to chain commands which generally should be avoided if there is no specific reason to do so.

giveen:
> Looks good so far. I assume the msi is these install wizards I remember right?
> How is this built is it like a Microsoft provided workflow? What is the licensing there can you please provide a link to the docs or similar?

https://docs.github.com/en/billing/concepts/product-billing/github-actions

internal/tool/implementations.go
giveen: No, not exactly by design in this pass.

The parsing/validation path is shared.
Linux/macOS can auto-run a small safe subset.
Windows currently requires confirmation for every shell command.
That stricter Windows behavior is intentional for safety first, then we can refine parity later.

internal/tool/implementations.go
giveen: PowerShell should remain the default for native Windows mode.

It aligns with Windows paths, user expectations, and shell semantics.
It avoids mixed-environment confusion (Windows filesystem + bash instructions).
Security/approval flow is still unified; only the shell backend differs.
Optional Bash-on-Windows support can be a later toggle.

internal/tool/implementations.go
giveen: Yes, this execution path is used on Windows too.

The tool path is shared cross-platform.
The process launcher is platform-specific underneath.
So wording should be platform-aware where model guidance is shown.


Agreed, and I’ll align that singular prompting

giveen:
@mlhher  https://github.com/mlhher/late/pull/19/changes/da7769042fe584952fad6391a63e70be28f96a14

internal/tool/implementations.go
mlhher: Please adjust the wording here. Plural is the correct as there is no specific amount set. We are giving a description of multiple commands and if the agent tried to run multiple commands at once this might poison context. For small models this will compound.

Also conceptually (gladly correct me) does this require changes to the example commands? Does pwsh also use "cat/echo > foo"? If not you should change these to platform-specific examples, specifically two so it can stay plural for all.

mlhher:
With the last change this will likely be ready to merge.

Thank you again for your work!

internal/tool/implementations.go
giveen: image


Confirmed that powershell does use 'echo/cat'

and switched back to the plurarl
https://github.com/mlhher/late/pull/19/changes/8bc0b4c4f3ff4e417e825cb051b577ba6e6674cd

mlhher:
I will test this a bit more, just to be safe and then this will be merged.

giveen:
> I will test this a bit more, just to be safe and then this will be merged.

Throw up a Windows VM and test it out would be my best suggestion.

mlhher:
Great work @giveen 
Thank you a lot again!

Issue #14, opened 2026-04-20
Streaming slows down beyond API tokens per second


phly95:
It seems that when I have over 20k tokens, the output seems to slow down a lot on my system. I'm running this on an older laptop while the AI is hosted on a newer gaming PC, which gets about 75 t/s with Qwen 3.6 35b a3b. Due to rendering slowness, the effective t/s is much, much slower when using late and scrolling becomes very slow as well.

It does seem to vary though. As sometimes it is fast. I guess there are some conditions where it slows down and others where it doesn't. Seems like some sort of text rendering bottleneck.

mlhher:
Please check if this happens more often when the content has more markdown content/syntax highlighting and generally formatted content.

You can reload sessions with the session subcommand or manually check inside ~/.local/share/late/sessions/

mlhher:
Fixed in `v1.0.4` as far as possible
This is caused by the rendering library. I aggressively optimized as much as possible for now should be faster with markdown heavy content but markdown content blocks will only receive styling after the respective block is closed.

Investigating further options. If this does not fix it for you re-open and provide a specific chat json where you experience this. If you do make sure to replace anything you might not want to share but keep the length.

PR #28, opened 2026-04-22
fix: TUI rendering in WSL terminals

diagonalfish:
### Description of Changes
Qwen3.6-generated fix - WSL terminals were failing to display the TUI entirely due to malformed ANSI escape sequences — the dark background color was prepended before lipgloss reset codes (\x1b[0m), producing invalid sequences that broke the terminal. Fix by injecting the background color *after* any full reset, and replace colored EOL fill with plain space padding.

- [x] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](./CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/internal/tui/view.go
+++ b/internal/tui/view.go
@@ -21,15 +21,22 @@ func (m Model) View() string {
 		m.statusBarView(),
 	)
 
-	// Fill each line's remaining width with our dark background using the
-	// terminal's Erase-in-Line sequence. This only processes ~40-50 lines
-	// (terminal height) so it's negligible — unlike the old appStyle.Render
-	// which parsed ANSI codes in the entire viewport content.
+	// Fill each line's remaining width with dark background.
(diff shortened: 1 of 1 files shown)


mlhher:
Thank you!
Please allow for some time to let me go over it manually. If there are no issues I will merge it.

diagonalfish:
No rush! Seems fine from what I've tried so far though.

diagonalfish:
Weirdly after other changes today this doesn't happen anymore on the current build, but I suspect there's still something weird going on with the ANSI codes that is occasionally causing WSL to barf. More study required, possibly.

mlhher:
Glad that this did not happen anymore.

We had some very similar logic in before but I explicitly removed it after #14. (Un-)Surprisingly this is rather costly and might slow down slower machines. I might evaluate how to proceed here in the long run.

For now if you hit any more issues please feel free to report it.

PR #30, opened 2026-04-22
feat(config): support OpenAI settings via app config

giveen:
- add openai_base_url, openai_api_key, and openai_model to app config

- centralize client config resolution with env > config > defaults precedence

- wire cmd/late startup through client.ResolveConfig

- return fallback defaults with surfaced errors on config create/parse failures

- add regression tests for config loading and resolver precedence

- document config keys, precedence behavior, and Windows config path

- [x] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](./CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ export OPENAI_BASE_URL="http://localhost:8080"
 # export OPENAI_MODEL="qwen3.5-35b-a3b"
 ```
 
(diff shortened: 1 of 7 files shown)


giveen:
Need to rebase first.

PR #47, opened 2026-04-26
fix(ast): enforce hard-block invariants in nested contexts

giveen:
## Summary
This PR applies a targeted safety fix to the AST-based shell analyzer and adds regression coverage.

## What we fixed
- Ensured hard-block patterns remain **hard-blocked in all AST contexts**, including nested constructs.
- Kept low-friction behavior for static allowlisted read-only commands.
- Added regression tests for nested blocked cases.

## Why we fixed it
The AST analyzer already blocked unsafe patterns like `cd` and output redirection, but nested control-flow/compound forms could degrade into confirmation-only behavior in some paths. That weakens the intended safety invariant.

## Changes
- `internal/tool/bash_analyzer.go`
  - Switched to a full AST walk in `Analyze()` so blocked patterns are detected even when nested.
- `internal/tool/bash_analyzer_test.go`
  - Added nested regression cases:
    - `if true; then cd /tmp; fi` -> blocked
    - `echo $(ls > out.txt)` -> blocked
- `internal/executor/executor.go`
  - Maintains fail-closed behavior for shell execution when explicit approval middleware is absent.

## Expected results
- These commands stay blocked (not just confirm):
  - `if true; then cd /tmp; fi`
  - `echo $(ls > out.txt)`
  - `( cd /tmp && ls )`
  - `while true; do echo x > f; break; done`
- Safe static commands are still low-friction (auto-allow):
  - `ls -la`
  - `git status --porcelain`
  - `grep -n foo file.txt`

## Validation
- `go test ./internal/tool/...` passed
- `go test ./...` passed

[X}By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository. (To check the box, put an x between the brackets like this: [x])

### DIFF
--- a/internal/executor/executor.go
+++ b/internal/executor/executor.go
@@ -81,16 +81,12 @@ func ExecuteToolCalls(ctx context.Context, sess *session.Session, toolCalls []cl
 		// Fail-closed: if no confirmation middleware is provided, do not
 		// execute shell commands (they must be explicitly approved by a
 		// middleware such as the TUI confirm middleware).
-		if len(middlewares) == 0 {
-			if t := sess.Registry.Get(tc.Function.Name); t != nil {
-				if _, ok := t.(*tool.ShellTool); ok {
-					result := "shell command requires explicit approval before execution"
-					if err := sess.AddToolResultMessage(tc.ID, result); err != nil {
-						return err
-					}
-					continue
-				}
+		if len(middlewares) == 0 && tc.Function.Name == "bash" {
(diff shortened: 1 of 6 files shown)


giveen:
@mlhher I have provided some recommended changes and fixes for windows. Other than that, I think the AST for bash is ready to go, and then I can start work on the WIndows side of the AST

mlhher:
Thanks, will take a look at it

PR #48, opened 2026-04-26
feat(approval): add once/session/project scopes with decay

giveen:
## Summary
Follow-up PR that introduces scoped approval decisions and automatic decay/revalidation to reduce prompt fatigue while preserving human-in-the-loop controls.

## What this adds
- Scoped confirmation choices:
  - `y` = allow once
  - `s` = allow session
  - `p` = allow project
  - `n` = deny
- Session-scoped approvals are kept in-memory and expire automatically.
- Project/global approvals are persisted with metadata and require re-validation after:
  - TTL expiry
  - app version change
- Backward-compatible loading for existing allow-list files.

## Why
This reduces repetitive prompts for clearly safe repeat actions without creating permanent trust that can silently drift over time.

## Key changes
- `internal/tool/permissions.go`
  - adds session approval storage + cleanup
  - adds TTL/version-aware metadata format for persisted approvals
  - keeps backward compatibility with old allow-list format
- `internal/tool/permissions_test.go`
  - tests session expiry
  - tests version/expiry filtering
  - tests metadata persistence/load
- TUI updates for new confirmation scope options:
  - `internal/tui/interactions.go`
  - `internal/tui/update.go`
  - `internal/tui/view.go`

## Validation
- `go test ./...` passed locally before this PR split.

## Target
- Head: `giveen/feat/approval-scopes-followup`
- Base: `mlhher/feat/ast_parser`

[X]By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository. (To check the box, put an x between the brackets like this: [x])

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 Standard AI coding assistants dump massive contexts into a single window, leading to token bloat, amnesia, hallucinations and degraded ability. **Late solves this by mirroring real engineering teams:** a Lead Architect orchestrator maps the codebase and spawns ephemeral, isolated subagents to execute perfect, exact-match code edits.
 
-![Late Orchestrator planning a 4-phase implementation and spawning the first subagent](assets/late-subagent-handoff.png)
-*Late acting as Lead Architect: Orchestrating a 4-phase plan and autonomously spawning atomic subagents.*
+![Late Orchestrator planning a multi-phase implementation and spawning the first subagent](assets/late-subagent-handoff.png)
+*Late acting as Lead Architect: Orchestrating a multi-phase plan and autonomously spawning atomic subagents.*
 
 > **Built with Late:** As of today, the vast majority of Late is being built *inside* Late.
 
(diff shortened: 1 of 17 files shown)


giveen:
@mlhher  this is a follow up to the AST, that will allow a bit of a better auto-approval process, and basically, if you are gone to long, it will force you to reconfirm permissions.

This is similar permission setup to a lot of coding IDE's like Visutal Studio Code.

mlhher:
I assume we can merge this separately. I will evaluate it. I want to keep complexity low though, I understand the reasoning but I am not sure if three separate allows are the right way for this. At least "Allow Always (Local/Project)" and "Allow Always (Global)" are easy to parse/understand.

giveen:
image


Allow Session = "since I started `late` this time", allow this command to run
Allow Project = "always allow this command in THIS project folder"
ALlow GLobal = "always allow this command, no matter the project"


The global is great for mcp commands

giveen:
stupid branch to branch PR and conflicts....grrr finally resolved.

mlhher:
I also see this prompt quite a bit now lol nice find on the mcp server.

I think my issue isn't necessarily with the commands it is more the wording and the potential to be not immediately obvious to the user.

I'd prefer having consistency more for all the "Allow". I see why you did it this way but there should be some more consistency and the label itself should just show it. For example "y" should be something like "Allow once". I get why you left out the "always" for the others it makes sense but I am worried this might introduce ambiguity to users.

giveen:
> I also see this prompt quite a bit now lol nice find on the mcp server.
> 
> I think my issue isn't necessarily with the commands it is more the wording and the potential to be not immediately obvious to the user.
> 
> I'd prefer having consistency more for all the "Allow". I see why you did it this way but there should be some more consistency and the label itself should just show it. For example "y" should be something like "Allow once". I get why you left out the "always" for the others it makes sense but I am worried this might introduce ambiguity to users.

Let me see what I can do to make it more "clear"

mlhher:
thanks! will take a look at it and then likely merge

giveen:
@mlhher  Okay fixed wording AND updated the quickstart guide to make it clear.

giveen:
And one last doc update that talks about the TTL decay feature of permissions.

mlhher:
Looks good.
The version lock might be a little too much as it is per minor version though I kind of do get it. Can leave it for now I think we can investigate this later (if at all).

Will merge this when ready (its on draft).

giveen:
Draft removed @mlhher

PR #51, opened 2026-04-28
AST-backed Windows PowerShell command analysis

giveen:
## AST-backed Windows PowerShell command analysis

This PR implements a full AST-backed analysis pipeline for PowerShell commands on Windows, replacing the existing string-matching `PowerShellAnalyzer` with a parse-tree driven approach backed by `System.Management.Automation.Language.Parser` — the same parser PowerShell itself uses.

The implementation ships with a gradual rollout system: legacy behavior is the default, shadow mode (log deltas, no behavior change) is opt-in via `LATE_AST_SHADOW=1`, and full enforcement via `LATE_AST_ENFORCEMENT=1`.

---

### `internal/tool/ast/` — new package

| File | Purpose |
|------|---------|
| `ir.go` | `ParsedIR` schema, `ReasonCode` constants, `Parser` interface |
| `ps_bridge.ps1` | Persistent PowerShell bridge process — parses via SMA.Language.Parser, emits line-delimited JSON IR |
| `windows_adapter.go` | Go side of bridge — singleton `bridgeProcess`, `roundTrip`, `Parse`, `CloseBridge` |
| `unix_adapter.go` | In-process Unix/bash adapter via `mvdan.cc/sh/v3` |
| `policy.go` | Platform-neutral `PolicyEngine.Decide(ir)` — all decisions derived from `ParsedIR` only |
| `shadow.go` | `ShadowAnalyzer` — runs AST in parallel with legacy, logs deltas, returns legacy result |
| `registry.go` | `NewParser(platform, cwd)` factory |
| `feature_flag.go` | `LATE_AST_SHADOW` / `LATE_AST_ENFORCEMENT` env var readers |

### `internal/tool/ast_bridge.go` — integration shim

Wires the `ast` package into `ShellTool.getAnalyzer()`. Routes between legacy, shadow, and enforcement modes based on feature flags. Seeds the Windows policy engine with the built-in safe cmdlet allow-list from `whitelistedWindowsCommands`.

---

### Persistent bridge process (not per-call spawn)

The initial approach spawned `pwsh.exe` per command (~700ms each). The final design uses a singleton `bridgeProcess` per Go process lifetime. Protocol: line-delimited JSON — `{"cmd":"..."}` request → one compact JSON IR response — bridge exits on stdin EOF. First call: ~700ms (amortised). All subsequent calls: ~1ms IPC round-trip.

### Fail-closed guarantee

Every code path that cannot produce a valid parse — sanitize failure, transport error, bridge crash, schema version mismatch, unknown risk flag — returns a `ParsedIR` with `ReasonSyntaxError` and a non-nil error. The policy engine and `astAnalyzer` both treat any error as `NeedsConfirmation=true`.

### Typed `ReasonCode` constants

All risk signals are typed constants validated on both sides of the transport. Unknown flags from the bridge are rejected with `ReasonSyntaxError`, preventing a compromised bridge script from silently suppressing risk signals.

### `ReasonDestructive` — semantically distinct from `ReasonInvokeExpr`

File-modification cmdlets (`Remove-Item`, `Copy-Item`, `Move-Item`, `Set-Content`, etc.) require user confirmation but are meaningfully different from dynamic-evaluation constructs (`Invoke-Expression`, `iex`). They get their own reason code and policy slot.

---

## Risk signal reference

| `ReasonCode` | Triggers | Policy outcome |
|---|---|---|
| `operator` | `\|`, `&&`, `\|\|`, `;` | Confirm unless all cmds allowlisted |
| `redirect` | `>`, `>>` to non-safe target | **Hard block** |
| `expansion` | `$var` (non-constant) | Confirm |
| `subshell` | `$(...)`, `(...)` | Confirm |
| `invoke_expression` | `iex`, `Invoke-Expression`, `-EncodedCommand` | Confirm |
| `cd` | `cd`, `Set-Location`, `Push-Location` | **Hard block** |
| `syntax_error` | Parse failure, transport error | Confirm (fail-closed) |
| `new_path` | `mkdir`, `New-Item` | Confirm (carveout: auto-approve if target is new path within cwd) |
| `destructive` | `Remove-Item`, `Copy-Item`, `Move-Item`, `Set-Content`, etc. | Confirm |

**Not flagged:** `$true`, `$false`, `$null`, `$_`, `$PSItem` — PowerShell language constants, filtered in the bridge before emitting `ReasonExpansion`. Avoids false-positive confirmation on `Where-Object { $_.Name -eq 'foo' }`.

**Not flagged:** `{ }` script-block expressions — idiomatic PS parameter syntax. `ScriptBlockExpressionAst` was deliberately excluded. Risk from dynamic execution is caught via `Invoke-Command` in `$invokeRiskCmdlets`, not the bare block node.

---

## Bugs fixed (vs. pre-PR baseline)

1. `write-host` missing from `whitelistedWindowsCommands` — prod/test divergence
2. `Remove-Item`/`Copy-Item` emitting `invoke_expression` — wrong semantic label
3. `gofmt` failure in `windows_adapter.go`
4. Goroutine blocked on timeout: stdin not closed → blocked goroutine until bridge process exit
5. No bridge shutdown path — no way to cleanly tear down the bridge
6. Error path used `Process.Kill()` directly — no stdin-close-first, leaving bridge in inconsistent state
7. Policy comment "Pure pipe" was wrong — rule covers all operators, not just pipes
8. `extractPowerShellTargetPath` didn't handle `New-Item -ItemType Directory -Path foo`
9. `isNewPath` `absPath` vs `canonicalPath` split was undocumented
10. `ScriptBlockExpressionAst` emitted `ReasonSubshell` — false-positive on all `Where-Object`/`ForEach-Object` blocks
11. `$true`/`$false`/`$null`/`$_` emitted `ReasonExpansion` — false-positive on constant variables
12. `snapshot_test.go` had no build tag — Windows corpus silently skipped on Linux CI
13. `startBridgeProcess` leaked stdin pipe on `StdoutPipe`/`Start()` failure
14. `TestWindowsParser_BridgeRestart` didn't `Wait()` the killed process
15. Feature flag routing (`getAnalyzer`) had no integration tests

---

### `internal/tool/ast/`

- `TestWindowsParser_BridgeContract` — 13 command patterns verifying exact risk flag contract
- `TestWindowsParser_SanitizeCommand` — null byte, length limit, boundary cases
- `TestWindowsParser_SanitizeCommand_IRShape` — sanitize failure produces valid IR
- `TestWindowsParser_ParseErrorShape` — bridge never panics on invalid input
- `TestWindowsParser_Concurrent` — 12 goroutines simultaneously, validates pipe protocol integrity
- `TestWindowsParser_BridgeRestart` — kills bridge externally, verifies auto-restart
- `TestWindowsCorpusSnapshot` _(Windows-only)_ — 22-entry corpus against full policy engine
- `TestPolicyEngine_Decide_SoftSignals` — extended to include `ReasonDestructive`

### `internal/tool/`

- `TestGetAnalyzer_NoFlags` — no flags → `*PowerShellAnalyzer`
- `TestGetAnalyzer_ShadowMode` — `LATE_AST_SHADOW=1` → `*shadowWrapper`
- `TestGetAnalyzer_EnforcementMode` — `LATE_AST_ENFORCEMENT=1` → `*astAnalyzer`
- `TestGetAnalyzer_EnforcementTakesPrecedence` — both flags set → `*astAnalyzer`
- `TestEnforcementMode_SafeCommandAutoApproves` — end-to-end: `Get-ChildItem` auto-approves
- `TestEnforcementMode_RiskyCommandRequiresConfirm` — `Remove-Item` confirm-only, not blocked
- `TestEnforcementMode_CdIsBlocked` — `cd` hard-blocked with non-nil `BlockReason`
- `TestEnforcementMode_ConstantVarNoConfirm` — `$true`/`$false`/`$null` auto-approve
- `TestShadowMode_ReturnsLegacyDecision` — shadow returns legacy result unchanged

---

## Known limitations (accepted, documented in code)

- **Windows Job Object:** if the Go process crashes (panic, `os.Exit`), the `pwsh.exe` bridge is orphaned. Proper fix requires `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` via `golang.org/x/sys/windows` — deferred to a follow-up.
- **`sc` alias collision:** `sc` in `$destructiveCmdlets` also matches `sc.exe` (Windows Service Control Manager). `sc query wuauserv` would require confirmation. Documented inline in `ps_bridge.ps1`.
- **`winPSPathOnce` singleton:** shell path is detected once per process; cannot be overridden in tests without process restart.

---

## Rollout

| Phase | Flag | Behavior |
|---|---|---|
| Default (now) | _(none)_ | Legacy `PowerShellAnalyzer`, no change for existing users |
| Shadow | `LATE_AST_SHADOW=1` | AST runs in parallel, deltas logged, legacy result returned |
| Enforcement | `LATE_AST_ENFORCEMENT=1` | AST pipeline is authoritative, legacy bypassed |

- [X ] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/internal/tool/ast/feature_flag.go
+++ b/internal/tool/ast/feature_flag.go
@@ -1,0 +1,31 @@
+package ast
+
+import "os"
+
+// Feature-flag environment variables for the AST rollout.
+// These are read at call-time so they can be toggled without restarting the
+// process (useful for integration tests and gradual rollout).
+const (
+	// EnvASTShadow enables shadow mode: the AST pipeline runs alongside the
+	// legacy analyzer and logs decision deltas but does NOT change behavior.
+	EnvASTShadow = "LATE_AST_SHADOW"
+
+	// EnvASTEnforcement promotes the AST pipeline to the authoritative path.
+	// When set, the legacy analyzer is bypassed entirely. This implies shadow
+	// mode as well (no need to set both).
+	EnvASTEnforcement = "LATE_AST_ENFORCEMENT"
+)
+
+// FeatureASTShadow reports whether AST shadow mode is enabled.
+func FeatureASTShadow() bool {
+	v := os.Getenv(EnvASTShadow)
+	return v == "1" || v == "true" || v == "on"
+}
+
+// FeatureASTEnforcement reports whether AST enforcement (Phase 5) is active.
+// When true, the AST policy path is authoritative and the legacy analyzer is
+// not consulted.
+func FeatureASTEnforcement() bool {
+	v := os.Getenv(EnvASTEnforcement)
+	return v == "1" || v == "true" || v == "on"
+}
--- a/internal/tool/ast/helpers.go
+++ b/internal/tool/ast/helpers.go
@@ -1,0 +1,51 @@
+package ast
+
+// addRiskFlag adds rc to ir.RiskFlags if not already present (idempotent).
+func addRiskFlag(ir *ParsedIR, seen map[ReasonCode]bool, rc ReasonCode) {
+	if !seen[rc] {
+		seen[rc] = true
+		ir.RiskFlags = append(ir.RiskFlags, rc)
+	}
+}
+
+// addStringUnique adds s to *slice if not already tracked in seen.
+func addStringUnique(slice *[]string, seen map[string]bool, s string) {
+	if !seen[s] {
+		seen[s] = true
+		*slice = append(*slice, s)
+	}
+}
+
+// appendUniqueRC returns a new slice with rc appended only if absent.
+func appendUniqueRC(sl []ReasonCode, rc ReasonCode) []ReasonCode {
+	for _, v := range sl {
+		if v == rc {
+			return sl
+		}
+	}
+	return append(sl, rc)
+}
+
+// hasRisk reports whether rc appears in ir.RiskFlags.
+func hasRisk(ir ParsedIR, rc ReasonCode) bool {
+	for _, r := range ir.RiskFlags {
+		if r == rc {
+			return true
+		}
+	}
+	return false
+}
+
+// emptyIR returns a correctly initialised ParsedIR with all slices non-nil.
+func emptyIR(platform Platform) ParsedIR {
+	return ParsedIR{
+		Version:     IRVersion,
+		Platform:    platform,
+		Commands:    []string{},
+		Operators:   []string{},
+		Redirects:   []string{},
+		Expansions:  []string{},
+		RiskFlags:   []ReasonCode{},
+		ParseErrors: []string{},
+	}
+}
--- a/internal/tool/ast/helpers.go
+++ b/internal/tool/ast/helpers.go
@@ -36,6 +36,16 @@ func hasRisk(ir ParsedIR, rc ReasonCode) bool {
 	return false
 }
 
+// HasRiskOnly reports whether the ONLY risk flag in ir is rc.
+// Exported so ast_bridge.go (in the tool package) can call it without
+// re-importing the full policy engine.
+func HasRiskOnly(ir ParsedIR, rc ReasonCode) bool {
+	if len(ir.RiskFlags) != 1 {
+		return false
+	}
+	return ir.RiskFlags[0] == rc
+}
+
 // emptyIR returns a correctly initialised ParsedIR with all slices non-nil.
 func emptyIR(platform Platform) ParsedIR {
 	return ParsedIR{
--- a/internal/tool/ast/helpers.go
+++ b/internal/tool/ast/helpers.go
@@ -57,5 +57,6 @@ func emptyIR(platform Platform) ParsedIR {
 		Expansions:  []string{},
 		RiskFlags:   []ReasonCode{},
 		ParseErrors: []string{},
+		CommandArgs: map[string][]string{},
 	}
 }
--- a/internal/tool/ast/ir.go
+++ b/internal/tool/ast/ir.go
@@ -57,12 +57,9 @@ const (
 	ReasonSyntaxError ReasonCode = "syntax_error"
 
 	// ReasonNewPath indicates the command creates a new path inside an allowed
-	// root. Used as an auto-approval carveout signal.
+	// root. The policy engine treats this as NeedsConfirmation; callers with
+	// cwd context may downgrade to auto-approve when the target is within scope.
 	ReasonNewPath ReasonCode = "new_path"
-
-	// ReasonAllowlisted indicates the command matched an explicit allow-list
-	// entry and was auto-approved.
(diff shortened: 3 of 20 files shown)


giveen:
@mlhher  All waiting for your review.

mlhher:
I'll take a look at it and test the Linux paths. Thanks

mlhher:
One issue:

With your new AST_ENFORCEMENT (which presumably should become the default to get rid of the legacy split) "find" with an "rm -rf" callback would slip through. The flags must be still checked we can never just rely on a command itself being ok and let all subcommands run. E.g. if I allow "find ..." it should not be allowed to run "find ... -exec" now. Similarly "curl google.com" should never also allow "curl github.com".

Please change that to align with the original behavior. Rest looks good will be merged afterwards.

@giveen

giveen:
@mlhher 

Sounds good , I'll get it done today.

giveen:
@mlhher 

 Fixed in 82d1c60.

**Root cause:** `allCommandsAllowlisted` only checked whether the command *name* appeared in the allow-list; it never looked at the stored flag set. So an approved `find .` would silently pass `find . -exec rm -rf {} \;`, and `curl google.com` would pass `curl github.com`.

**What changed:**

- Added `CommandArgs map[string][]string` to `ParsedIR` — populated by both the Unix adapter (walking `CallExpr` args) and the PowerShell bridge (walking `CommandAst` elements), with flags normalized the same way the legacy `BashAnalyzer` did (`--flag=value → --flag`, `-N → -*`).
- `allCommandsAllowlisted` now validates that every flag actually present in the invocation (`ir.CommandArgs[cmd]`) is also in the stored allow-list for that command. Any flag not in the stored set fails the check and falls through to `NeedsConfirmation` — exactly the original behaviour.
- Five new sub-tests in `TestPolicyEngine_FlagEnforcement`, including an end-to-end `find . -exec rm -rf {} \;` case parsed through `UnixParser`.

mlhher:
Looks good.

TestUnixCorpusSnapshot fails on the two ls examples because of this change, please fix these and it'll be merged

giveen:
@mlhher 
Fixed in 9d519b2.

The snapshot corpus was seeding `ls` with an empty flag map (`{}`), so `ls -la` and `ls -rt` started failing once `allCommandsAllowlisted` began validating flags. Updated the seeded `AllowedCommands` to include `-la` and `-rt` (matching what a real user would have stored after approving those invocations), and removed the now-stale comment claiming flag-level validation wasn't performed by the AST path.

PR #53, opened 2026-04-29
feat(context-rot): add --context-rot flag with safe pruning and 400-fix

giveen:
### Background

Long-running agentic sessions accumulate chat history until the context window fills up, after which the LLM silently starts losing earlier turns or the API begins rejecting requests. This PR adds an **opt-in** mechanism to detect that condition and recover automatically — trimming history to a clean window, re-injecting the mission plan from disk, and continuing without losing the current task.

---

### The Problem

Two compounding issues were discovered and fixed during development:

#### 1. `status: 400` errors from OpenAI-compatible APIs mid-recovery

The original heartbeat position fired **between** `AddAssistantMessageWithTools` (which commits the assistant's `tool_calls` entry to history) and `ExecuteToolCalls` (which commits the corresponding `tool` result messages). If pruning happened at that point, the rebuilt history tail could contain the assistant's `tool_calls` message but none of its results — or neither — while `ExecuteToolCalls` would then append tool results referencing `tool_call_id`s that no longer existed in history. OpenAI-compatible APIs treat this as a malformed message structure and return `status: 400`.

#### 2. History tail window splitting tool exchanges

The 10-message tail window can cleanly bisect a multi-turn tool exchange near the boundary. An assistant message with `tool_calls` may be retained in the tail while its `tool` result messages are cut off (they fell outside the window), producing the same `400` on the very next request.

---

#### `--context-rot` flag &mdash; `cmd/late/main.go`

New opt-in CLI flag. **Off by default** — existing sessions are completely unaffected.

```sh
late --context-rot
```

Wires into `Session.ContextRecoveryEnabled`. Two signals trigger recovery:

| Signal | Condition |
|---|---|
| **Token overflow** | Prompt token usage > 75% of the known context window |
| **Message ceiling** | History exceeds 80 messages (deterministic fallback for backends that don't report token counts) |

---

#### `PruneAndRestoreFromDisk()` rewrite &mdash; `internal/session/session.go`

Full rewrite of the recovery function with five discrete steps:

1. **Tail extraction** — capture the last 10 messages as a snapshot
2. **First boundary guard** — trim leading non-`user` messages so the tail always starts on a clean turn boundary
3. **Structural sanitizer** (`sanitizeTailToolCalls`) — scan the tail for any `assistant` message whose `tool_calls` IDs are not all resolved by `tool` result messages also present in the tail; drop the offending assistant message and any orphaned tool messages that follow it
4. **Second boundary guard** — re-apply after sanitization in case removing an unresolved turn exposed a new non-`user` head
5. **Mission re-injection** — if `implementation_plan.md` exists in the working directory and is under 8 KB, prepend it as a `user` message so the LLM retains its mission after the reset

---

#### Split heartbeat &mdash; `internal/executor/executor.go`

The context-recovery check was split into two safe positions in `RunLoop`:

- **Position A** — immediately before the no-tool-calls early return: history is fully consistent (no pending tool calls), safe to prune
- **Position B** — immediately after `ExecuteToolCalls` returns: all tool results for this turn are committed to history before the tail is rebuilt

This eliminates the race window between assistant tool-call commit and tool-result commit where pruning would produce orphaned references.

---

#### Orchestrator fixes &mdash; `internal/orchestrator/base.go`

Two bugs fixed in `Execute()` uncovered during testing:

- **TUI status desync** — a `defer "idle"` status event was firing *after* the `"closed"` event, causing the TUI to flip back to "Ready" while the orchestrator was still active. Removed the defer; all status transitions are now explicit.
- **Empty user message injection** — subagents call `Execute("")` after their goal is already injected into history. The missing guard was adding a blank `user` message on every subagent turn, gradually corrupting history structure. Fixed with `if text != ""`.

---

#### Mandatory read-before-edit rule &mdash; `internal/assets/prompts/instruction-coding.md`

Added a `## MANDATORY: Always Read Before You Edit` section to the coder subagent system prompt. The root cause of a separate loop bug was the coder copying code blocks from the architect's goal description directly into `target_edit` search blocks — those snippets are often stale or reformatted. The rule requires a `read_file` call before any edit, no exceptions.

---

### Test Coverage

**`internal/session/prune_test.go`** — 12 unit tests:

- History reduction
- System prompt immutability
- Boundary guard (tail must start with `user`)
- Mission injection — present / absent / oversized (> 8 KB)
- Small and empty history edge cases
- Persistence sync verification
- Signal threshold arithmetic
- **`TestPrune_OrphanedToolCallsRemoved`** — directly exercises the sanitizer with a crafted history where the 10-message window bisects a tool exchange; asserts zero unresolved `tool_call_id`s remain after pruning

**`internal/executor/executor_test.go`** — two new `RunLoop` tests using a live `httptest` SSE server:

- `TestRunLoop_ContextRecovery_Disabled` — with flag off, history grows past 80 messages without pruning
- `TestRunLoop_ContextRecovery_Enabled` — with flag on and an 81-message prefill, history is pruned back to ≤ 10 messages after a single turn

All 171 existing tests continue to pass.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,7 @@ install: build ## Build and install the binary to your Go bin path
 
 run: build ## Build and run the project
 	@./bin/${BINARY_NAME}
+
+build-test: ## Build the late-test performance evaluation harness
+	@echo "Building late-test..."
+	@go build -o bin/late-test ./cmd/late-test
--- a/cmd/late-test/main.go
+++ b/cmd/late-test/main.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -208,7 +209,7 @@ func main() {
 
 	// ── Session ───────────────────────────────────────────────────────────────
 	const systemPrompt = "You are a helpful coding assistant. Answer concisely."
-	histPath := fmt.Sprintf("/tmp/late-test-session-%d.json", time.Now().UnixMilli())
+	histPath := filepath.Join(os.TempDir(), fmt.Sprintf("late-test-session-%d.json", time.Now().UnixMilli()))
 	history := make([]client.ChatMessage, 0, *prefill)
 
 	if *prefill > 0 {
--- a/cmd/late-test/main.go
+++ b/cmd/late-test/main.go
@@ -1,0 +1,397 @@
+// late-test is a manual performance evaluation harness for the context-recovery
+// architecture. It connects to the real LLM configured in config.json, pre-fills
+// the session history to a configurable depth to simulate context pressure, then
+// runs an interactive loop while printing per-turn recovery metrics.
+//
+// Usage:
+//
+//	./bin/late-test [flags]
+//
+// Flags:
+//
+//	-prefill  N     Pre-fill the session with N synthetic messages before starting
+//	                (default 0; use >=81 to immediately trigger the 80-msg ceiling)
+//	-prompt   text  Seed prompt to send as the first real turn (default: built-in)
+//	-turns    N     Maximum number of interactive turns before the harness exits
+//	                (default 0 = unlimited)
+//	-plan           Write a synthetic implementation_plan.md before starting,
+//	                so the plan-injection path is exercised (default true)
+package main
+
+import (
+	"bufio"
+	"context"
+	"flag"
+	"fmt"
+	"math"
+	"os"
+	"strings"
+	"time"
+
+	appconfig "late/internal/config"
+	"late/internal/client"
+	"late/internal/executor"
+	"late/internal/session"
+)
+
+// ── ANSI helpers ──────────────────────────────────────────────────────────────
+
+const (
+	reset  = "\033[0m"
+	bold   = "\033[1m"
+	cyan   = "\033[36m"
+	yellow = "\033[33m"
+	green  = "\033[32m"
+	red    = "\033[31m"
+	grey   = "\033[90m"
+)
+
+func header(s string) string { return bold + cyan + s + reset }
+func warn(s string) string   { return yellow + s + reset }
+func ok(s string) string     { return green + s + reset }
+func bad(s string) string    { return red + s + reset }
+func dim(s string) string    { return grey + s + reset }
+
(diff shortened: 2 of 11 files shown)


mlhher:
Thanks I'll test this and try it out

mlhher:
I am unsure about this.
The bad error request is good no doubt there that can get merged basically now.

The context rot flag I am unsure honestly. I do get the reason but the auto compressing really is something I dont like. What is your thought vs the new context limits I pushed to main? These will warn you (or outright stop you) either if you send or if the agent is in an agentic loop.

giveen:
This may be bit too ambitious, and I"m not a fan of summarization either. However, if you dont like this, completely okay.  I plan on looking at native integration of https://github.com/mksglu/context-mode next.

and I do like the warnings we have now, we are working with local models, not endless 1M context frontier cloud models.

And maybe the user just needs to take a break lol
@mlhher

mlhher:
The issue I see here is the automatic summarization of the architect. That will degrade reasoning and should not happen automatically. I am fine if it is provided as an explicit command that the user has to trigger maybe you can integrate it with the new popup that shows up when we near context limits/llama.cpp throws the error which we now parse? Also fine to close it out (maybe something for a command palette?). Really the only issue I have is it happening without user intervention.

I'll leave it open for now, I am open to ideas here.

giveen:
For me, it was an experiment in ideas that i spent a couple days talking to AI about various ways to improve long term context storage, which is why i added in its own -context-rot flag to try out. Its hard to say 100% over a long term coding session if it truly helped or not.

mlhher:
The flag is a fair point as long as it's explicitly opt in.

Let's go the way of us testing it for a tiny bit. If both of us find it helps it will be merged. I do like the idea as it could potentially solve the issue long-term.

If you do notice anything please feel free to share your findings I'd be interested too (I have a hard time hitting context limits in realistic use cases which may bias me here slightly).

Though if we do come to merge it you would need to adjust the context warnings as I assume it would be a false warning now; not sure how you implemented it currently but if not happening this could turn instead into a notif like "hey we're summarizing now ctx limits hit"

late
diagonalfish, line 1: Committed the binary

giveen:
I'm going to close this out as a dead idea.

PR #62, opened 2026-05-06
cleanup: remove legacy shell analyzers and AST rollout flags

giveen:
## Summary
This PR completes the AST analyzer migration by removing legacy rollout infrastructure, shadow mode scaffolding, and hardcoded analyzer fallbacks. The codebase now enforces a single, deterministic AST-based command analysis path across all platforms.

**Net Result**: ~1,243 lines of dead rollout code removed, ~271 lines of focused utility code added. All tests passing. Zero regressions detected.

---

## Motivation

After successfully implementing AST analyzers for both Unix and Windows platforms, the legacy analyzer implementations and feature flag infrastructure became dead code:

1. **Legacy Bash Analyzer** (Unix fallback) — Only reachable when feature flags were unset; with always-on AST, this became unreachable.
2. **Legacy PowerShell Analyzer** (Windows legacy) — Fully migrated to AST; never instantiated on Windows anymore.
3. **Shadow Mode (Phase 4 monitoring)** — Existed to compare AST vs legacy decisions during migration; served no runtime purpose once both analyzers were feature-equivalent.
4. **Feature Flags** (`LATE_AST_SHADOW`, `LATE_AST_ENFORCEMENT`) — Controlled rollout behavior; no longer needed with single code path.

Keeping this code creates:
- **Maintenance burden**: Two parallel implementations of the same policy logic to keep in sync.
- **Risk of silent divergence**: Untested fallback paths could bitrot without catching it.
- **Cognitive overhead**: Future contributors must reason about multiple analyzer branches.
- **Test pollution**: Tests for unused rollout modes add noise without coverage value.

This cleanup eliminates all that debt while retaining the allow-list parsing utility that the permission system still needs.

---

### Phase 1: Remove Shadow Mode Infrastructure
- **Deleted** `internal/tool/ast/shadow.go` (88 lines) — Entire shadow analyzer and monitoring logic.
- **Modified** `internal/tool/ast_bridge.go` — Removed `shadowAnalyzerShim` and `shadowWrapper` adapter structs.
- **Modified** `internal/tool/ast_mode_test.go` — Removed `TestShadowMode_ReturnsLegacyDecision` test.

### Phase 2: Remove Feature Flags & Collapse Analyzer Selection
- **Deleted** `internal/tool/ast/feature_flag.go` (31 lines) — Removed `LATE_AST_SHADOW`, `LATE_AST_ENFORCEMENT` constants and feature functions.
[...]

PR #63, opened 2026-05-06
## feat: optional session archive compaction (rag-doll)

giveen:
eat: optional session archive compaction (rag-doll)

Summary
- Adds an opt-in rolling session archive system that moves older messages out of
  the active context window into a compressed on-disk archive, provides keyword
  search and stable message retrieval, and exposes two new tools for model use.

What this does
- When a session grows beyond a configured threshold, Late moves the oldest
  messages into a `.archive.json` file next to the session history and trims the
  active history window.
- Adds two tools:
  - `search_session_archive` — keyword search across archived messages, returns
    ranked results with stable reference handles.
  - `retrieve_archived_message` — fetch full archived messages by reference.
- Emergency compaction: if a turn hits the context window limit mid-stream, an
  emergency compaction runs, the turn is retried, and the session remains usable.
- Subagents inherit the parent orchestrator's archive so they can search full
  session history as well.
- Opt-in only: zero behavior change without the config block.

Files/areas to review
- `internal/archive/*` (archive/compaction/search)
- `internal/tool/archive_tools.go`
- `internal/orchestrator/*` (archive integration / subagent inheritance)
- `internal/session/*` (archive metadata and save/load hooks)

Removed from this PR (moved to separate branches / backups)
- Several later "correctness/audit" commits were removed from this PR to keep
  review focused on the compaction feature. Those fixes (audit passes, metadata
  fixes, UI hides, list filtering, etc.) are split into separate branches so
  they can be reviewed independently:
  - `split/allowlist` — allowlist / whitelist changes
  - `split/subagent` — planner / subagent changes
  - `split/session-listing` — `ListSessions` filtering fix
- Backups of the previous states were created:
  - `backup/orig-rag-doll` (original PR state backup)
  - `backup/rag-doll-clean-backup` (cleaned/alternate backup)

Notes
- This PR was restored to the original compaction feature head (commit `0488ddc`)
  to make review focused and incremental.

- [X] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/cmd/late/main.go
+++ b/cmd/late/main.go
@@ -226,6 +226,14 @@ func main() {
 		sess.Registry.Register(t)
 	}
 
+	// Log archive compaction startup status (Phase 7 bootstrap).
+	if appConfig != nil && appConfig.IsArchiveCompactionEnabled() {
+		settings := appConfig.ArchiveCompactionSettings()
+		fmt.Fprintf(os.Stderr, "[late] archive compaction enabled (threshold=%d, keepRecent=%d)\n",
+			settings.CompactionThresholdMessages, settings.KeepRecentMessages)
+		_ = tool.RegisterArchiveTools // referenced to ensure linkage; actual registration done at pre-run hook
+	}
+
 	// Initialize common renderer
 	renderer, _ := glamour.NewTermRenderer(
 		glamour.WithStylesFromJSONBytes(tui.LateTheme),
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -1,0 +1,224 @@
+// Package archive provides session archive persistence, compaction, and search.
+// It is dependency-free (stdlib + late/internal/client only) so both internal/session
+// and internal/tool can import it without creating a cycle.
+package archive
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"late/internal/client"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+const SchemaVersion = 1
+
+// ArchivePath derives the archive file path from a history path.
+// If historyPath ends in ".json", replaces suffix; otherwise appends.
+func ArchivePath(historyPath string) string {
+	if strings.HasSuffix(historyPath, ".json") {
+		return strings.TrimSuffix(historyPath, ".json") + ".archive.json"
+	}
+	return historyPath + ".archive.json"
+}
+
+// LockPath derives the lock file path from a history path.
+func LockPath(historyPath string) string {
+	if strings.HasSuffix(historyPath, ".json") {
+		return strings.TrimSuffix(historyPath, ".json") + ".archive.lock"
+	}
+	return historyPath + ".archive.lock"
+}
+
+// BaseSessionID extracts the session ID token from a history file path.
+// e.g. "/sessions/session-abc.json" → "session-abc"
+func BaseSessionID(historyPath string) string {
+	base := filepath.Base(historyPath)
+	if strings.HasSuffix(base, ".json") {
+		return strings.TrimSuffix(base, ".json")
+	}
+	return base
+}
+
+// HashMessage returns a stable sha256 hex hash of a ChatMessage's JSON representation.
+func HashMessage(msg client.ChatMessage) string {
+	data, _ := json.Marshal(msg)
+	sum := sha256.Sum256(data)
+	return fmt.Sprintf("%x", sum)
+}
+
+// HashBytes returns a sha256 checksum of raw bytes.
+func HashBytes(data []byte) [32]byte {
+	return sha256.Sum256(data)
+}
+
+// SessionArchive is the top-level on-disk archive structure.
+type SessionArchive struct {
+	SessionID            string         `json:"session_id"`
+	SchemaVersion        int            `json:"schema_version"`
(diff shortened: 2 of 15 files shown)


giveen:
Its a psuedo-RAG system, basically.

mlhher:
I will go over this after #62 is merged

mlhher:
Might take a bit to fully test and evaluate this

giveen:
> Might take a bit to fully test and evaluate this

No worries, its not going anywhere, just trying to solve the issue everyone has with local LLM's and context size.

mlhher:
@giveen Can you remove the unrelated changes we might look at them separately but having multiple things stuffed into this pr is making it rather hard to evaluate it in isolation. Looking over the code what I can see on a quick glance:

planner subagent
whitelists
nil map
session listing

and anything i didnt catch. if any of them are required for the branch obviously leave it in but these do seem unrelated to me.

giveen:
okay think i got it all

mlhher:
Please also the code in ast_bridge. I get what you did that but this may require some more discussion (maybe not; but I'd rather prefer if we can have it separated in case it does require some more thought; I do get why you did it though and I agree it should be done somehow).

Code looks good from what I got so far. Will need to try this out and see how it behaves in real usage.

giveen:
Reverted changes in `internal/tool/ast_bridge.go` — summary:

- Removed (the PR additions):
  - `whitelistedUnixCommands` (new Unix-safe command list).
  - Expanded `whitelistedWindowsCommands` (new map of Windows cmdlets/aliases).
  - Seeding `allowed` for both Unix and Windows (switch that added Unix seeding).
  - Analyze() change that unconditionally auto-approved `ReasonNewPath` (unsupervised mode) without checking the target path or cwd.
  - Deletion of legacy shims (`shadowAnalyzerShim`, `shadowWrapper`).

- Restored (original behavior brought back):
  - Only seed Windows built-ins into `allowed` (no automatic Unix seeding).
  - Reinstated the "new-path carveout" that checks the target path against `cwd` before auto-approving:
      if d.NeedsConfirmation && !d.IsBlocked && ir.Platform == ast.PlatformWindows {
          if ast.HasRiskOnly(ir, ast.ReasonNewPath) {
              if target := extractPowerShellTargetPath(command); target != "" && isNewPath(target, a.cwd) {
                  return CommandAnalysis{NeedsConfirmation: false}
              }
          }
      }
    (i.e., auto-approve mkdir/New-Item only when the target is within the session cwd on Windows)
  - Restored `shadowAnalyzerShim` and `shadowWrapper` to bridge `ast.LegacyAnalysis`/`ast.ShadowAnalyzer` to `CommandAnalyzer`.
@mlhher

mlhher:
@giveen Can you share your findings how does it feel like (and which model)? I have a hard time hitting limits at 65k with my workflow and I do not want to artificially boggle it down as that would not give me a realistic feel for it.

giveen:
> @giveen Can you share your findings how does it feel like (and which model)? I have a hard time hitting limits at 65k with my workflow and I do not want to artificially boggle it down as that would not give me a realistic feel for it.

I often start off looking at code with a "Explain this codebase" prompt to my models. This starts with a deep dive of basically reading everything and then giving me a summary.

I recently switched to [Qwopus3.5-9B-Coder](https://huggingface.co/Jackrong/Qwopus3.5-9B-Coder-GGUF) as my planner with a 256k context and then DeepSeekv4 Flash (cloud) as the subagent due to limitations.

Keeping track of the session allows it to keep a psuedo memory of what happened in the past and why a change was done.

I have always agreed that summirezation is a context loss, but being able to read back on a recent change and why it was changed helps...a bit.

mlhher:
> I have always agreed that summirezation is a context loss, but being able to read back on a recent change and why it was changed helps...a bit.

I agree that having it is better than just not having it. I will need to try this out a little before merging it, I just want to see how it feels like. Don't take not merging it right now as not wanting it, just need some time to fully evaluate and I am quite short on time currently.

> I recently switched to [Qwopus3.5-9B-Coder](https://huggingface.co/Jackrong/Qwopus3.5-9B-Coder-GGUF) as my planner with a 256k context and then DeepSeekv4 Flash (cloud) as the subagent due to limitations.

Interesting, I was about to say that the reasoning of at least Claude and Gemini is clearly hidden but I see that they acknowledge that. Might have to check that out though considering it is dense I likely assume that 35B will provide similar speeds. Will check it out thanks for linking!

mlhher:
@giveen I think you need to update some things, the content type got changed from string to a typed struct due to the multi-modal content.

```go
if len(full) != 1 || full[0].Content
```

Also archiveSub, runArchivePreHook and some of the ast_bridge stuff (whitelistedWindowsCommands, extractPowerShellTargetPath, ast.LegacyAnalysis) seems undefined not sure what happened there.

giveen:
@mlhher I havent seen any real improvements with this branch over a long term sessions across multiple work sessions. 
I believe I am killing this idea.

mlhher:
This has been brought up often enough for me to want to implement it. I rely on your evaluations here since this is your PR.

giveen:
The idea is sound but my tooling of it is not sustainable. I think it still gets lost and ends up resulting in a long session run failure.

PR #69, opened 2026-05-19
Nesw feature : Orchestrator's conversation compression

Sarki:
### Description of Changes

Blunt implementation of the automatic compression of the Orchestrator conversation, highly inspired by Codex's approach.

**An LLM was used in the development of this feature, all changes have been thoroughly reviewed, tested as much as possible and fixed when appropriate by a human before proposing this change.**


Sarki:
Alright, after more "ground" testing, obviously the term "blunt" is accurate but it works.

Issues/enhancements I've found : 
* In the PR I've removed `CompressionThreshold` in the method `defaultConfig` from `config.go` which may interfere with the trigger of the compression, reintroducing it with a default value of `0` might be better.
* After a compression event the Orchestrator will often rewrite `implementation_plan.md` and ask for a new approval.
    * This is actually not a bad thing per-se, maybe adding a  new flag to always accept the given implementation after a compression event? (two `compression-summary-prefix.md` files? or revise how the compression reorganises the conversation?)
    * This is interesting in a good way because the compression events forces the Orchestrator to review what has been done so far and sometimes it catches errors/false assumptions made by the subagents before. Alas it may be frustrating for long running tasks (as in tedious and laborious ones).

mlhher:
Thank you for your contribution. Please also check out #63 as it seems these are both attempts to solve the same issue. If you feel either approach has advantages, feel free to elaborate and we can evaluate both.

Sarki:
Thanks for your time, after multiple retries on the same long running task, I've figured out that this way (compressing, or perhaps the way I implemented it) is not as efficient as hoped on the long run and pushes toward exactly what is described in #63 (which is some quite impressive work as well).

I'm dropping the PR, I'll propose my actual finding which apparently works for tedious, repetitive and long running tasks (almost) like a charm. But I wouldn't call this compression...

PR #74, opened 2026-05-20
fix: don't double-append /v1 when BaseURL already contains a path prefix

### LINKED ISSUE #72: `openai_base_url` incompatible with providers that use non-standard API path prefixes (e.g. z.ai, Azure)
`openai_base_url` in `config.json` does not support non-standard API path prefixes because late-cli unconditionally appends `/v1/chat/completions` to whatever base URL is configured.

This means providers that use a versioned or namespaced base path (e.g. `https://api.z.ai/api/coding/paas/v4`) end up with a doubled path:

https://api.z.ai/api/coding/paas/v4/v1/chat/completions

which returns a `404`.

**Expected behavior:** `openai_base_url` is treated as a true base, and the final endpoint is constructed as `{openai_base_url}/chat/completions` — consistent with how the official OpenAI SDK handles base URLs.

**Actual behavior:** `/v1` is always appended, making the setting incompatible with any provider whose base URL already contains a path prefix.

**Workaround:** run a local reverse proxy that accepts requests on a plain base URL and rewrites the path to the provider's actual endpoint before forwarding.

HenrikSondergaard:
## Problem

`openai_base_url` in `config.json` does not support non-standard API path prefixes because `internal/client/client.go` unconditionally appends `/v1/chat/completions` to whatever base URL is configured:

```go
url := strings.TrimSuffix(c.cfg.BaseURL, "/") + "/v1/chat/completions"
```

Providers that use a versioned or namespaced base path (e.g. `https://api.z.ai/api/coding/paas/v4`) end up with a doubled path:

```
https://api.z.ai/api/coding/paas/v4/v1/chat/completions  → 404
```

## Fix

Introduce a `chatCompletionURL()` helper that parses the base URL and applies the following logic — consistent with the [OpenAI Python SDK convention](https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py) that `base_url` is a true base the caller controls:

- If the base URL has **no path** (e.g. `http://localhost:8080`), `/v1` is appended automatically for backwards compatibility.
- If the base URL **already has a path** (e.g. `https://api.z.ai/api/coding/paas/v4`), it is used as-is and only `/chat/completions` is appended.

```go
func (c *Client) chatCompletionURL() string {
    base := strings.TrimSuffix(c.cfg.BaseURL, "/")
    if u, err := url.Parse(base); err == nil && (u.Path == "" || u.Path == "/") {
        base += "/v1"
    }
    return base + "/chat/completions"
}
```

This correctly handles all common cases:

| `openai_base_url` | Result |
|---|---|
| `http://localhost:8080` | `http://localhost:8080/v1/chat/completions` ✓ |
| `http://localhost:8080/v1` | `http://localhost:8080/v1/chat/completions` ✓ |
| `https://api.z.ai/api/coding/paas/v4` | `https://api.z.ai/api/coding/paas/v4/chat/completions` ✓ |

Closes #72

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -50,18 +50,6 @@ func NewClient(cfg Config) *Client {
 	}
 }
 
-// chatCompletionURL builds the chat completions endpoint from BaseURL.
-// It strips a trailing /v1 from BaseURL before appending /v1/chat/completions,
-// so both plain bases (http://localhost:8080) and bases that already include
-// the version segment (http://localhost:8080/v1) produce the correct URL,
-// and providers with a non-standard path prefix (https://api.z.ai/api/coding/paas/v4)
-// are not broken by a spurious extra /v1.
-func (c *Client) chatCompletionURL() string {
-	base := strings.TrimSuffix(c.cfg.BaseURL, "/")
-	base = strings.TrimSuffix(base, "/v1")
-	return base + "/v1/chat/completions"
-}
-
 // ChatCompletion sends a chat prompt to the OpenAI-compatible endpoint.
 func (c *Client) ChatCompletion(ctx context.Context, req ChatCompletionRequest) (*ChatCompletionResponse, error) {
 	if c.getBackend() == BackendUnknown || (c.getBackend() == BackendLlamaCPP && c.ContextSize() == -1) {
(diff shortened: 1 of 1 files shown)


mlhher:
Thank you for your contribution.

Similarly with the PR template here:
(Or you can use the template)

For the code: I agree this is the right way, though as you noted some people will likely already rely on our automatic parsing. For that case specifically it would be better if you can additionally add some safety fallback like this:

```go

    base := strings.TrimSuffix(baseURL, "/")
    if u.Path == "" || u.Path == "/" {
        base += "/v1"
    }
    return base + "/chat/completions", nil
```

that way people who are using a bare URL will just have their v1 appended automatically.

HenrikSondergaard:

mlhher:
Thank you again! Both look good and will be merged later on today.

PR #78, opened 2026-06-02
fix: namespace allowed_tools.json entries as {server}:{tool}

### LINKED ISSUE #77: allowed_tools.json entries should be namespaced by MCP server to prevent tool name collisions
## Problem

`allowed_tools.json` stores trust entries using only the tool name as key (e.g. `"list_files"`). If two MCP servers expose a tool with the same name, there is no way to distinguish which server the trust was granted for — the agent may auto-approve the wrong one, or the file becomes ambiguous about what was actually approved.

## Proposed fix

Namespace entries as `{server}:{tool}`, for example:

```json
"graph-rag:list_files": { ... }
"github:get_file_contents": { ... }
"mssql:execute_sql": { ... }
```

This matches the convention already used by tools like Claude Code (`mcp__server__tool`) and makes the file human-readable without needing any additional documentation.

Old single-name entries could be treated as unscoped/legacy and prompt for re-approval on next use, ensuring a clean migration path.

## Benefit

- Eliminates tool name collision between MCP servers
- Makes `allowed_tools.json` self-documenting — the entry tells you exactly which server and tool was approved
- Consistent with established MCP tool naming conventions

HenrikSondergaard:
## Problem

`allowed_tools.json` stores trust entries using only the bare tool name as the key (e.g. `"list_files"`). If two MCP servers expose a tool with the same name, there is no way to distinguish which server the trust was granted for — the agent may silently auto-approve the wrong one.

## Fix

Four files changed: `internal/mcp/client.go` (core), `cmd/late/main.go` (backward-compat), `cmd/mcp-run/main.go` (call-site update), and a new `internal/mcp/client_test.go`.

`ToolAdapter` gains a `serverName` field, populated from the config map key in `ConnectFromConfig`. `Name()` now returns `"{server}:{tool}"` (e.g. `"graph-rag:list_files"`):

```go
func (t *ToolAdapter) Name() string {
    if t.serverName != "" {
        return t.serverName + ":" + t.mcpTool.Name
    }
    return t.mcpTool.Name
}
```

A `BareName()` method is also added to expose the original MCP tool name for callers that need it.

Because `Name()` drives both tool registration and `tc.Function.Name`, all permission checks and saves in the TUI middleware (`LoadAllAllowedTools`, `SaveAllowedTool`, `SaveSessionAllowedTool`) automatically use the namespaced key — no changes needed in the permission system.

`Connect()` now accepts a `serverName string` argument, and sessions are stored by server name instead of a hardcoded `"default"` key — fixing a pre-existing bug where `Close()` would only shut down the last connected MCP subprocess.

The `enabledTools` check in `cmd/late/main.go` now checks the namespaced name first, then falls back to the bare name, so existing configs that disable tools by bare name (e.g. `"list_files": false`) continue to work without modification.

`cmd/mcp-run/main.go` is updated to pass `"default"` as the server name for ad-hoc single-server connections.

## Behaviour after this change

| `openai_base_url` | allowed_tools.json key |
|---|---|
| Before | `"list_files"` |
| After | `"graph-rag:list_files"` |

This matches the established convention used by Claude Code (`mcp__server__tool`) and makes the file self-documenting.

## Migration

Legacy unscoped entries in existing `allowed_tools.json` files will not match the new namespaced names, so users are prompted for re-approval on next use — a clean one-time migration with no data loss.

Closes #77

### DIFF
--- a/cmd/late/main.go
+++ b/cmd/late/main.go
@@ -217,10 +217,24 @@ func main() {
 	sess := session.New(c, historyPath, history, systemPrompt, *useToolsReq)
 	executor.RegisterTools(sess.Registry, enabledTools, true)
 
-	// Register MCP tools into the session registry
+	// Register MCP tools into the session registry.
+	// MCP tool names are now namespaced as "{server}:{tool}" (e.g. "graph-rag:list_files").
+	// For backwards compatibility with configs that disable tools by bare name
+	// (e.g. "list_files": false), we check the namespaced name first, then fall
+	// back to the bare name so existing configs keep working without modification.
 	for _, t := range mcpClient.GetTools() {
-		// MCP tools are enabled by default unless explicitly set to false in config.json
-		if enabled, exists := enabledTools[t.Name()]; exists && !enabled {
+		name := t.Name()
+		// Derive the bare name by stripping the server prefix.
+		bareName := name
+		if idx := strings.LastIndex(name, ":"); idx >= 0 {
+			bareName = name[idx+1:]
+		}
+		// Namespaced entry takes priority over bare-name entry.
+		if enabled, exists := enabledTools[name]; exists {
+			if !enabled {
+				continue
+			}
+		} else if enabled, exists := enabledTools[bareName]; exists && !enabled {
 			continue
 		}
 		sess.Registry.Register(t)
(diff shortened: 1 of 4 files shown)


HenrikSondergaard:

PR #79, opened 2026-06-02
fix: raise MCP truncation limit to 32,768 chars and fix UTF-8 slicing

### LINKED ISSUE #76: MCP tool results hard-capped at 4000 bytes with no config knob — truncates normal issue/PR/search responses
MCP tool results are truncated to a hardcoded **4000 bytes** in `internal/mcp/client.go:81-84` (`ToolAdapter.Execute`). Built-in `Read`/`Bash` tools allow **32768 bytes** (`internal/tool/implementations.go:231,234`) — so MCP results are capped **8× more strictly** than built-in tools. Many ordinary MCP responses (GitHub issue bodies, PR diffs, search results) exceed 4000 bytes and get cut mid-string with the marker `[... truncated, output exceeded limit ...]`.

```go
// internal/mcp/client.go
// Truncate to ~1k tokens (~4000 chars)
const maxChars = 4000
if len(output) > maxChars {
    output = output[:maxChars] + "\n\n[... truncated, output exceeded limit ...]"
}
```

## Reproduction

Call `mcp__github__get_issue` on an issue whose body is ~2,400 chars via a real GitHub MCP server. The combined response (title + metadata + body) passes 4000 bytes, so the result is cut mid-word. Deterministic and reproducible — same cut point every time. The same MCP call returns the full body in Claude Code, confirming this is a late-cli-side cap, not an MCP-server or model-API limit.

## Problems

1. **Limit far lower than built-in tools** — 4000 vs 32768 bytes.
2. **Not configurable** — `maxChars` is a `const`. There is no env var, CLI flag (`cmd/late/main.go`), or config key (`internal/config/config.go` `Config` struct) to raise it.
3. **Byte-based slice** — `output[:maxChars]` cuts on a byte boundary and can split a multi-byte UTF-8 rune.
4. **Misleading comment** — the comment says "~1k tokens (~4000 chars)", but `len()` on a Go string counts **bytes**, not chars or tokens.
5. **No test coverage** — no test exercises this truncation path (only the built-in `Read`/`Bash` paths at 32768 are tested in `internal/tool/implementations_test.go`).

## Requested fix (any of)

- **(a)** Raise the default substantially — e.g. match the 32768 already used for built-in tools.
- **(b)** Per-tool / per-server override so individual MCP tools can opt into larger limits.
- **(c)** Raw-passthrough mode for MCP results (no truncation).

Additionally, regardless of which option: make the limit configurable (CLI flag + config key), slice on rune boundaries instead of bytes, fix the misleading comment, and add a test covering the MCP truncation path.

## Affected code

- `internal/mcp/client.go:81-84` — the cap and marker
- `internal/tool/implementations.go:231,234` — built-in limits (32768) for comparison

HenrikSondergaard:
## Problem

The `Execute` method on `ToolAdapter` (`internal/mcp/client.go`) hard-caps MCP tool responses at 4,000 bytes while the built-in tools (`read_file`, `bash`) use 32,768 chars (defined in `internal/tool/implementations.go:231-234`). This 8x discrepancy causes legitimate MCP responses (GitHub issues, PR diffs, search results) to be truncated mid-string.

The current 4,000-byte cap on MCP tool responses actually hurts Late's context discipline rather than helping it. Truncated partial data wastes context tokens on garbled information the model can't act on, often triggering retries that further pollute the context window — the exact noise Late's architecture is designed to eliminate. The built-in tools already established 32,768 as the correct limit; MCP tools should match since they surface the same kind of data through a different transport. This fix also corrects a UTF-8 bug where byte-based slicing could split multi-byte characters.

## Changes (`internal/mcp/client.go`)

1. **Raise `maxChars` from 4000 to 32768** to match the built-in tool limit.
2. **Fix UTF-8 bug**: `output[:maxChars]` sliced by bytes, which can corrupt multi-byte characters (e.g. Japanese, emoji). Now converts to `[]rune` before slicing and back to `string` after:
   ```go
   runes := []rune(output)
   if len(runes) > maxChars {
       output = string(runes[:maxChars]) + "\n\n[... truncated, output exceeded limit ...]"
   }
   ```
3. **Fix the comment**: said `"~4000 chars"` but `len()` counts bytes. Updated to accurately describe what the limit does.

## Tests (`internal/mcp/client_test.go`)

Four cases covering the assignment requirements:

- Output under the limit passes through unchanged
- Output at exactly the limit passes through unchanged
- Output over the limit is truncated with the marker appended
- Multi-byte UTF-8 characters (Japanese `日`, 3 bytes each) at the boundary are not split — the test checks for `utf8.RuneError` (`'�'`) in the output

Closes #76

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -77,10 +77,13 @@ func (t *ToolAdapter) Execute(ctx context.Context, args json.RawMessage) (string
 
 	output := sb.String()
 
-	// Truncate to ~1k tokens (~4000 chars)
-	const maxChars = 4000
-	if len(output) > maxChars {
-		output = output[:maxChars] + "\n\n[... truncated, output exceeded limit ...]"
+	// Truncate to 32,768 Unicode characters — matching the built-in tool limit
+	// (maxReadFileChars / maxBashOutputChars in internal/tool/implementations.go).
+	// Slicing by rune rather than byte prevents splitting multi-byte UTF-8 sequences.
+	const maxChars = 32768
+	runes := []rune(output)
+	if len(runes) > maxChars {
+		output = string(runes[:maxChars]) + "\n\n[... truncated, output exceeded limit ...]"
 	}
 
 	return output, nil
--- a/internal/mcp/client_test.go
+++ b/internal/mcp/client_test.go
@@ -1,0 +1,75 @@
+package mcp
+
+import (
+	"strings"
+	"testing"
+)
+
+// truncateOutput replicates the truncation logic in ToolAdapter.Execute so it
+// can be tested independently without a live MCP server.
+func truncateOutput(output string) string {
+	const maxChars = 32768
+	runes := []rune(output)
+	if len(runes) > maxChars {
+		return string(runes[:maxChars]) + "\n\n[... truncated, output exceeded limit ...]"
+	}
+	return output
+}
+
+func TestTruncateOutput_UnderLimit(t *testing.T) {
+	input := strings.Repeat("a", 100)
+	got := truncateOutput(input)
+	if got != input {
+		t.Errorf("expected output to pass through unchanged, got len=%d", len(got))
+	}
+}
+
+func TestTruncateOutput_AtLimit(t *testing.T) {
+	input := strings.Repeat("a", 32768)
+	got := truncateOutput(input)
+	if got != input {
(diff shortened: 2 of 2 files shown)


HenrikSondergaard:

mlhher:
Thank you!

PR #39, opened 2026-04-24
Constrain shell auto-approval to new paths within cwd

giveen:
## PR Title
Harden shell approval path handling and reduce unnecessary Windows confirmations

## Summary
This PR narrows scope to two focused areas:

1. **Path safety hardening for shell auto-approval**
2. **Windows-specific confirmation policy improvements**

It intentionally excludes unrelated work (subagent init changes and Unix native-first shell execution routing).

---

### 1) Path hardening for shell auto-approval
- Tightens auto-approval so it applies only to safe, constrained creation flows.
- Improves path handling to avoid approval bypass via ambiguous or expansion-prone targets.
- Keeps approval conservative when command intent is unclear.

Key outcomes:
- Safer handling of path-based command approval decisions.
- Reduced risk of approving commands that rely on shell expansion/ambiguity.

---

### 2) Windows confirmation policy: less noisy, still conservative
- Replaces “always prompt” behavior with **selective auto-approval** for clearly safe/read-only PowerShell commands.
- Introduces risky-command/syntax checks that still force explicit confirmation.
- Adds parser-backed command classification for more reliable decision-making.
- Supports simple new-path creation carveouts where policy allows.

Key outcomes:
- Fewer unnecessary prompts on Windows for benign commands.
- Maintains explicit confirmation for risky/mutating or unclear command shapes.

---

## Scope Clarification
This PR **does not** include:
- Subagent lazy init / retry changes
- Unix native-first execution/fallback refactor
- Temporary tracing instrumentation used during local verification

---

## Files Changed
- `internal/tool/implementations.go`
- `internal/tool/permissions.go`
- `internal/tool/implementations_test.go`
- `internal/tool/implementations_bash_test.go`
- `internal/tool/implementations_cmd_test.go`

---

### Automated
- `go test ./internal/tool` ✅
- `GOOS=windows GOARCH=amd64 go test -c ./internal/tool` ✅ (Windows compile-path validation)

### Behavioral checks covered by tests
- Path auto-approval hardening behavior
- Expansion/ambiguous target rejection in auto-approval extraction
- Windows selective confirmation behavior
- Windows parser-backed command extraction/classification
- Windows new-path carveout behavior

---

## Why This Matters
- Improves shell safety posture without over-broad prompt fatigue.
- Keeps approval policy predictable and conservative.
- Improves day-to-day UX on Windows while preserving explicit confirmation boundaries.

PR #83, opened 2026-06-21
feat: add search_tool as a native alternative to bash grep/find

giveen:
### Description of Changes

- Add new SearchTool (internal/tool/search.go + search_test.go) that walks directories with filepath.WalkDir, reads files line-by-line, and matches patterns with Go stdlib regexp — no external dependencies
- Register search_tool alongside read_file, bash, etc. in executor.go, gated by the enabled_tools config map
- Enable search_tool by default in config.go, with a backward-compatible merge that fills in missing defaults for existing config files
- Fix broken indentation in executor.go shell-tool middleware check (was nested inside an orphan if-block, breaking the fail-closed guard)
- Update instruction-coding.md and instruction-planning.md to recommend search_tool over bash grep/find

- [ X] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/internal/assets/prompts/instruction-coding.md
+++ b/internal/assets/prompts/instruction-coding.md
@@ -11,8 +11,7 @@ Your goal is defined by the main agent. You are typically asked to write code, r
 - You should use `write_file` or `target_edit` to modify code as instructed.
 - You should evaluate whether to use `write_file` or `target_edit` based on the context.
 - You must prefer native tools (e.g. `search_tool`, `write_file`, and `target_edit`) over bash commands (e.g. `grep`, `find`, `echo`, and `sed`).
-- **PREFERRED**: Use `search_tool` for finding files or searching file contents instead of `bash` with `grep`, `find`, or `rg`. `search_tool` returns structured {path, line, content} matches, respects permission gates, and applies per-tool output caps for efficient results.
-- If you use `bash` for a search command (`grep`, `rg`, `find`), the bash gate will refuse the command and remind you to use `search_tool`.
+- If you use `bash` for a search command (e.g. `grep`, `rg`, `find`), the tool will refuse your command and remind you to use the `search_tool` instead.
 
 ## Ambiguity
 
--- a/internal/assets/prompts/instruction-coding.md
+++ b/internal/assets/prompts/instruction-coding.md
@@ -11,7 +11,8 @@ Your goal is defined by the main agent. You are typically asked to write code, r
 - You should use `write_file` or `target_edit` to modify code as instructed.
 - You should evaluate whether to use `write_file` or `target_edit` based on the context.
 - You must prefer native tools (e.g. `search_tool`, `write_file`, and `target_edit`) over bash commands (e.g. `grep`, `find`, `echo`, and `sed`).
-- Use `search_tool` for finding files or searching file contents instead of `bash` with `grep` or `find`.
+- **PREFERRED**: Use `search_tool` for finding files or searching file contents instead of `bash` with `grep`, `find`, or `rg`. `search_tool` returns structured {path, line, content} matches, respects permission gates, and applies per-tool output caps for efficient results.
+- If you use `bash` for a search command (`grep`, `rg`, `find`), the bash gate will refuse the command and remind you to use `search_tool`.
 
 ## Ambiguity
 
--- a/internal/assets/prompts/instruction-coding.md
+++ b/internal/assets/prompts/instruction-coding.md
@@ -1,24 +1,30 @@
 You are a **Coding Subagent** invoked by a main agent to perform specific coding tasks.
 
 ## Goal
+
 Your goal is defined by the main agent. You are typically asked to write code, refactor functions, or fix bugs in specific files.
 
 ## Capabilities
+
 - You have access to the same tools as the main agent, **IN ADDITION** you also have access to file-modifying tools (`write_file`, `target_edit`) that are withheld from the main agent.
 - You should use `read_file` to understand the context.
 - You should use `write_file` or `target_edit` to modify code as instructed.
 - You should evaluate whether to use `write_file` or `target_edit` based on the context.
-- You must prefer native tools (e.g. `write_file` and `target_edit`) over bash commands (e.g. `echo` and `sed`).
+- You must prefer native tools (e.g. `search_tool`, `write_file`, and `target_edit`) over bash commands (e.g. `grep`, `find`, `echo`, and `sed`).
+- Use `search_tool` for finding files or searching file contents instead of `bash` with `grep` or `find`.
 
 ## Ambiguity
+
 - If you encounter any issue or ambiguity you must immediately stop with your implementation.
 - Instead you must report back to the main agent a summary of what you have changed so far together with the exact issues or ambiguities you have encountered.
 - After encountering issues do not try to identify or solve the issue on your own. The main agent will solve it for you as long as you give it proper context.
 
 ## Current working dir
+
 Your current working directory is `${{CWD}}`
 
 ## Output
+
 - When you have completed your coding task, report back to the main agent.
 - Confirm exactly what changes you made.
 - If you encounter any issues or have to deviate from the plan or there are ambiguities, immediately stop whatever you are doing and return to the main agent with an explanation of what you have done so far and what the issues are.
--- a/internal/assets/prompts/instruction-planning.md
+++ b/internal/assets/prompts/instruction-planning.md
@@ -6,8 +6,8 @@ Your goal is to analyze complex user requests, explore the existing codebase to 
 
 **CRITICAL: You are an ARCHITECT, not a CODER.**
 
-* **YOU CAN**: Read files, search the codebase with `search_tool` (preferred over bash+grep/find/rg), list directories, and analyze project structure.
-* **SEARCH PREFERENCE**: For code search and pattern matching, always use `search_tool` instead of `bash` with `grep`, `rg`, or `find`. `search_tool` returns structured results and respects permission boundaries. The bash gate will block search commands in `bash` and redirect you to `search_tool`.
+* **YOU CAN**: Read files, search the codebase with `search_tool`, list directories, and analyze project structure.
+* **YOU MUST**: Use the `search_tool` instead of using the `bash_tool` with e.g. `grep`/`find`/`rg` to search for and match patterns and strings in the codebase.
 * **YOU MUST**: Use `write_implementation_plan` to record your design before any execution.
 * **YOU MUST**: Use `spawn_subagent` (type `coder`) for **ALL** direct file modifications. **CRITICAL TOOL RULE: You MUST invoke the `spawn_subagent` tool MULTIPLE TIMES—exactly once for EVERY individual step in your Implementation Plan. You are strictly forbidden from passing multiple steps or the entire plan into a single `spawn_subagent` call.**
 * **YOU CANNOT**: Edit files, create files (other than the plan), or run destructive bash commands.
--- a/internal/assets/prompts/instruction-planning.md
+++ b/internal/assets/prompts/instruction-planning.md
@@ -3,49 +3,61 @@ You are the **Lead Architect and Planning Agent**.
 Your goal is to analyze complex user requests, explore the existing codebase to understand the context, and generate a rigorous, step-by-step **Implementation Plan**.
 
 ## 1. Capabilities & Restrictions
+
 **CRITICAL: You are an ARCHITECT, not a CODER.**
 
-*   **YOU CAN**: Read files, search the codebase, list directories, and analyze project structure.
-*   **YOU MUST**: Use `write_implementation_plan` to record your design before any execution.
-*   **YOU MUST**: Use `spawn_subagent` (type `coder`) for **ALL** direct file modifications. **CRITICAL TOOL RULE: You MUST invoke the `spawn_subagent` tool MULTIPLE TIMES—exactly once for EVERY individual step in your Implementation Plan. You are strictly forbidden from passing multiple steps or the entire plan into a single `spawn_subagent` call.**
-*   **YOU CANNOT**: Edit files, create files (other than the plan), or run destructive bash commands.
-    *   *Note: Direct file-editing tools (like `write_file` or `target_edit`) are physically removed from your toolset. You MUST delegate all coding to subagents.*
-    *   *Even for requests to "implement", "add", "update", or "edit", you MUST follow the plan -> subagent pipeline. Direct edits are only for subagents.*
+* **YOU CAN**: Read files, search the codebase with `search_tool`, list directories, and analyze project structure.
+* **YOU MUST**: Use `write_implementation_plan` to record your design before any execution.
+* **YOU MUST**: Use `spawn_subagent` (type `coder`) for **ALL** direct file modifications. **CRITICAL TOOL RULE: You MUST invoke the `spawn_subagent` tool MULTIPLE TIMES—exactly once for EVERY individual step in your Implementation Plan. You are strictly forbidden from passing multiple steps or the entire plan into a single `spawn_subagent` call.**
+* **YOU CANNOT**: Edit files, create files (other than the plan), or run destructive bash commands.
+  * *Note: Direct file-editing tools (like `write_file` or `target_edit`) are physically removed from your toolset. You MUST delegate all coding to subagents.*
+  * *Even for requests to "implement", "add", "update", or "edit", you MUST follow the plan -> subagent pipeline. Direct edits are only for subagents.*
 
 ## 2. Your Workflow
+
 You must not just "guess" the plan. You must **investigate** first to ensure your plan is grounded in reality. If an `AGENTS.md` exists make sure to read it first.
 
 ### Phase 1: Exploration & Discovery
+
 Before proposing a plan, you must gather information.
-1.  **Map the Geography**: Understand the project structure if unknown.
-2.  **Trace the Logic**: Find relevant code patterns or specific string occurrences, and read files to examine the content of specific files.
-3.  **Identify Constraints**: Look for existing patterns (e.g., "all API responses use `ApiResponse` struct") and ensure your plan adheres to them.
+
+1. **Map the Geography**: Understand the project structure if unknown.
+2. **Trace the Logic**: Find relevant code patterns or specific string occurrences, and read files to examine the content of specific files.
+3. **Identify Constraints**: Look for existing patterns (e.g., "all API responses use `ApiResponse` struct") and ensure your plan adheres to them.
 
 ### Phase 2: Strategic Thinking
+
 Construct a mental model of the solution. Ask yourself:
-*   What files need to be modified?
-*   What new files need to be created?
-*   How can this be broken down into atomic, verifiable steps?
-*   Are there any **Agent Skills** (e.g., brand guidelines, specialized tools) that either you or the subagents should activate?
+
+* What files need to be modified?
+* What new files need to be created?
+* How can this be broken down into atomic, verifiable steps?
+* Are there any **Agent Skills** (e.g., brand guidelines, specialized tools) that either you or the subagents should activate?
 
 ### Phase 3: Architectural Stress Test & Conflict Resolution
+
 Before generating the final output, you must internally simulate the execution of your plan.
+
 1. **Contradiction Check**: Does any step in Phase 2 directly conflict with a rule established in Phase 1? (e.g., removing a parameter but adding a CLI flag for it later).
 2. **I/O & Memory Sanity**: Are you requesting the system to load massive amounts of data just to read a small subset? If so, specify the exact memory-efficient parsing method.
 3. **Concurrency Safety**: If touching files, state explicitly *when* a lock is acquired and *when* it is released to prevent deadlocks.
 
 ### Phase 4: Deliver the Plan
+
 Output a structured **Implementation Plan** in Markdown. This plan will be handed off to an *Execution Agent* (a junior developer AI) who will follow your instructions blindly. Clarity and precision are paramount.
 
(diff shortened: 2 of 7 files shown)


mlhher:
Can you explain your reasoning for this, I am curious. I remember couple months ago I explicitly removed the search tool I had implemented because it seemed like most models were preferring Bash over the native tool during testing.

Note that I have only skimmed the code so far.

giveen:
> Can you explain your reasoning for this, I am curious. I remember couple months ago I explicitly removed the search tool I had implemented because it seemed like most models were preferring Bash over the native tool during testing.
> 
> Note that I have only skimmed the code so far.

to be honest, i wasnt aware that you had done that or I just forgot, i was just working on late and then saw that it didnt have a native search so i thought it was odd. I tested with Qwen3-Coder-Next and also Deepseek-v4-flash and both called up the search_tool with no issues. Let me try with Qwen3.6 and Gemma4 models and see if they have any issues.

I'll also look at your previous search tool and see whats different and why I havent had an issue using my search_tool.

giveen:
@mlhher 

Initial anlysis, still need to test:
# search_tool Analysis: Re-adding a Search Tool After Removing One Upstream

## Authorship

| Who | Repo | Action |
|-----|------|--------|
| **you (giveen)** | `giveen/late` (this fork) | Added `search_tool` (commit `ed24f69`) |
| **you (giveen)** | `mlhher/late-cli` (upstream) | Authored PR #39 that **removed** `grep_search` |
| **mlhher** | `mlhher/late-cli` (upstream) | Maintainer who merged the removal |

So you were on both sides of this coin — you removed `grep_search` upstream, then added `search_tool` in your own fork. This is interesting because it means the upstream's reasons for removal are *your own past reasoning*, and now you want to understand why you're doing the opposite.

---

## Part 1: What Was Removed Upstream (PR #39, commit `22b16fe`)

The upstream `GrepTool` was named `grep_search`:

```go
type GrepTool struct{}
func (t GrepTool) Name() string        → "grep_search"
func (t GrepTool) Description() string → "Search for a pattern in files within a directory"
```

Parameters: `query`, `path`, `case_sensitive`, `is_regex`

**How it worked:**
- Walked directories with `filepath.WalkDir`
- Pure-Go line matching (no shell execution)
- **Bucketed by top-level directory** for diversity
- Capped at **10 results** (`maxGrepResults = 10`)
- Showed only **first match per file**, no context
- No glob filtering, no output modes, no content view

**Why PR #39 removed it (per commit message):**
> *"fix: make bash tool more robust, remove grep tool"*

The replacement: an upgraded `BashTool` with:
1. A **command whitelist** (`grep`, `find`, `ls`, `cat`, `head`, `tail`, etc.)
2. **Arg-level security** — rejects `;`, `|`, `&`, `..`
3. **Path safety** — `IsSafePath()` enforces CWD boundaries
4. **Output limits** — 1024 lines, 32K chars
5. **Fallback parsing** — if the command field contains spaces, split into command+args

The reasoning (from the upstream conversation in PR #39, April 24-25):

> **mlhher:** *"I am currently thinking of overhauling it entirely tbh."*
> ...
> **mlhher:** *"I want to try and test an AST parser. It would solve all (most) issues."*
> ...
> **mlhher:** *"So I have a more or less working AST solution... This means likely going forward we will use your logic for Windows and the AST parser for bash."*

The bash tool later evolved to use `mvdan/sh` for AST-level command parsing, making `grep` through bash fully safe and analyzable.

---

## Part 2: The "LLMs Prefer Grep" Problem

The core insight is about **LLM behavior**, not tool quality:

1. **LLMs are `grep` natives** — their training data contains millions of shell commands with `grep -r`, `find -name`, etc. When searching code, an LLM instinctively reaches for `bash { "command": "grep -r 'pattern' ." }`.

2. **A separate search tool creates friction** — the LLM must:
   - Notice it among many other tools in the list
   - Trust an unfamiliar tool over one it's used millions of times
   - Convert its grep knowledge into a different parameter schema
   - Remember the specific name (`grep_search` or `search_tool` vs. just `grep`)

3. **Prompt engineering is fragile** — telling the LLM "prefer X over Y" works in some turns but not others. In a multi-turn session, the LLM may default back to `bash + grep` when context pressure increases.

The upstream's solution: **don't fight it.** Make `bash + grep` safe via whitelist + AST analysis, remove the competing tool entirely, and let the LLM use what it naturally prefers.

---

## Part 3: What Our `search_tool` Is vs. What the Old `grep_search` Was

This is the critical difference. The old `grep_search` was **weak**. Our `search_tool` is **much stronger**.

| Feature | Old `grep_search` (removed) | Our `search_tool` |
|---------|----------------------|-------------------|
| Regex | ✅ | ✅ |
| Case sensitivity | ✅ | ✅ `case_sensitive` |
| Literal/fixed_strings | ❌ | ✅ `fixed_strings` |
| Glob/include filter | ❌ | ✅ `include: "*.go"` |
| Output modes | ❌ (only file+first line) | ✅ `files_with_matches`, `content`, `count` |
| Context lines | ❌ | ✅ `context_lines` (before/after) |
| Max results | Fixed at **10** | ✅ configurable, default 100, **up to 500** |
| Binary detection | ❌ | ✅ null-byte check |
| VCS/vendor skip | ✅ (.git, node_modules, vendor) | ✅ (.git, node_modules, .svn, .hg) |
| Hidden file skip | ❌ | ✅ |
| Long line truncation | ❌ | ✅ at 1000 chars |
| Output char cap | ❌ | ✅ at 32,768 chars |
| Context cancellation | ❌ | ✅ |
| Empty file handling | ❌ | ✅ |
| **Description** | "Search for a pattern in files within a directory" | "Search files and file contents using regex or literal patterns. Returns matching files and/or content with line numbers. **Use this instead of bash grep/find.**" |
| **Name** | `grep_search` (LLM could confuse with grep) | `search_tool` (clearly a tool, not grep) |

Our `search_tool` is a **substantially better tool** than the old `grep_search`. The old one was barely more useful than a quick `grep -rl` — capped at 10 results with no way to see content. Ours is a proper code search tool with context lines, glob filtering, and multiple output modes.

**Important context:** The bash tool in our fork already has the AST analysis inherited from upstream. So `bash { "command": "grep -r 'pattern' ." }` is already safe. `search_tool` is competing against a **secure** `bash + grep`, not a wild-west shell.

---

## Part 4: The Real Comparison — search_tool vs. Secure bash+grep

Since bash is already AST-secured, the real question is: **what value does `search_tool` add beyond what the LLM can already do with `bash + grep`?**

### Things `search_tool` does better:

| Use Case | `bash + grep` | `search_tool` |
|----------|--------------|---------------|
| Find all Go files with a function | `grep -rl "func " --include="*.go" .` | `search_tool(pattern="func ", include="*.go")` |
| See matching lines with context | `grep -rn -C 3 "pattern" .` (verbose flags LLM must remember) | `search_tool(pattern="pattern", output_mode="content", context_lines=3)` |
| Count matches per file | `grep -rc "pattern" .` (grep -c counts files, not total matches) | `search_tool(pattern="pattern", output_mode="count")` |
| Case-insensitive literal search | `grep -ri "pattern" .` | `search_tool(pattern="pattern", fixed_strings=true)` (no regex escaping needed) |
| Skip vendor/node_modules | Must know to use `--exclude-dir` or `find ... -path ... -prune` | Automatic |
| Search multiple extensions | Must pass multiple `--include` flags | `search_tool(pattern="pat", include="*.{go,ts,rs}")` |

### Things `bash + grep` still does that `search_tool` can't:
- Inverted matches (`grep -v`)
- Word boundary matching (`grep -w`)
- Line number only output (`grep -n`)
- Custom regex flags (`-P` for Perl-compatible regex)
- Binary file handling choices (`-I`, `-a`)
- Pipe to other commands for further processing

### The LLM preference problem:
Even with a better tool, the LLM may still default to `bash + grep` because:
- `grep` is deeply embedded in its training data
- `bash + grep` is one fewer tool to discover and learn
- Under token pressure, the LLM reaches for familiar patterns
- Prompt instructions are not guarantees

---

## Part 5: Why Add It Back Now?

The old `grep_search` was removed because:

1. **It was weak** — 10 results, no context, no output modes, no filtering → LLMs had no reason to use it over `bash + grep`
2. **`bash + grep` was being made safe** — AST analysis meant grep through bash was acceptable
3. **Code complexity** — ~150 lines of Go for a tool nobody used

Our `search_tool` is different because:

1. **It's genuinely more capable** — context lines, glob filtering, content mode with proper formatting, count mode, literal mode. It does things `grep -r` can't easily express.
2. **It handles edge cases** — binary detection, long line truncation, VCS skip, hidden files — all things the LLM would need to remember to pass as flags to grep
3. **It has a consistent interface** — one tool, one expected output format vs. differences between `grep`, `rg`, `ag`, `find` across systems
4. **The description explicitly tells the LLM to use it** — "Use this instead of bash grep/find" (the old description didn't have this)
5. **The name avoids confusion** — `search_tool` looks like a native tool (`read_file`, `write_file`, `target_edit`), not a shell command

---

## Summary

| | Old `grep_search` | Our `search_tool` |
|---|---|---|
| **Capability** | Weak — 10 results, no context | Rich — 500 results, context, filters, modes |
| **Name** | `grep_search` (confusable with grep) | `search_tool` (clearly a tool) |
| **Description** | Generic | Tells LLM to use it |
| **Bash safety** | Bash was being AST-secured | Bash is already AST-secured |
| **Competition** | Loses to bash+grep on features | **Wins** on structured output, edge cases |
| **LLM adoption risk** | Low (LLM prefers grep) | Medium (LLM may still prefer grep, but tool is better when used) |

The old `grep_search` lost because it was **worse** than `bash + grep` in every way. Our `search_tool` might win because it's **better** in specific ways — but adoption is still an LLM behavior problem, not a tool quality problem. The upstream's lesson was that **you can't prompt-engineer LLMs away from grep**. Our counter-argument is: give them a tool that's good enough that they **want** to use it.

giveen:
# search_tool Improvements

## Overview

We added a comprehensive `search_tool` as a native alternative to `bash grep/find/rg` for code search, then improved it with a multi-layered strategy to make LLMs **prefer it** over shell commands. The approach combines grep-mapped parameters, a "bash gate" that blocks search-shaped shell commands, better tool descriptions, prompt reinforcement, and a performance fix.

---

## Phase 1: Grep-Mapped Parameters

**What**: Made `search_tool` parameters map naturally to grep concepts so the LLM recognizes them.

**Files changed**: `internal/tool/search.go`

### Changes

| Feature | grep equivalent | Description |
|---------|----------------|-------------|
| `output_mode: "matching-files"` | `grep -l` | Alias for `files_with_matches` |
| `output_mode: "text"` | `grep -n` | Alias for `content` |
| `exclude: "*.min.js"` | `grep --exclude` | Glob pattern to skip files |
| `recursive: false` | `grep` without `-r` | Non-recursive search (default: true) |

### Why
LLMs think in grep's language. The old enum `["files_with_matches", "content", "count"]]` meant nothing to a model trained on `grep -l`, `grep -n`, `grep -c`. Now it can use what it knows.

---

## Phase 2: Bash Gate

**What**: When the LLM tries to use `bash` with a search command (`grep`, `rg`, `find`, `ag`, `ack`, `fd`), it gets a structured error pointing to `search_tool`.

**Files changed**: `internal/tool/implementations.go`

### Blocking Logic

| Command | Blocked? | Why |
|---------|----------|-----|
| `grep -rn "foo" .` | ✅ **Blocked** | grep is always a search tool |
| `rg "pattern"` | ✅ **Blocked** | ripgrep is always a search tool |
| `find . -name "*.go"` | ✅ **Blocked** | `-name` flag indicates search intent |
| `find . -iname "*.ts"` | ✅ **Blocked** | `-iname` flag indicates search intent |
| `find . -path "*/test/*"` | ✅ **Blocked** | `-path` flag indicates search intent |
| `find . -type f -exec chmod 644 {} \;` | ❌ **Allowed** | No name/path flag; it's a file operation |
| `find . -size +1M -delete` | ❌ **Allowed** | No name/path flag; it's a file operation |
| `cat file \| grep foo` | ❌ **Allowed** | First word is `cat`, not a search command |
| `ls` / `cat` / `pwd` | ❌ **Allowed** | Not search commands |

### Configuration

The gate is configurable via the `LATE_BASH_GATE` environment variable:

| Value | Behavior |
|-------|----------|
| `enforce` (default) | Blocks search commands with error + tool hint |
| `warn` | Logs warning but allows execution through |
| `allow` | Disables the gate entirely |

### Error Message
When blocked, the LLM sees:
```
bash refused: `grep` detected. Use the native `search_tool` instead — it returns
structured {path, line, content} matches, respects permission gates, and applies
per-tool output caps.
```

### Why
Research shows description hints alone are "marginal" (measured 15/27 bash preference even with "PREFERRED" hints). The bash gate creates a **negative feedback loop**: the LLM tries the familiar path, gets blocked, and learns to use `search_tool` instead.

---

## Phase 3: Better Tool Descriptions

**What**: Strengthened search_tool and bash descriptions with explicit preference hints.

**Files changed**: `internal/tool/search.go`, `internal/tool/implementations.go`

### search_tool Description (before)
```
Search files and file contents using regex or literal patterns.
Returns matching files and/or content with line numbers.
Use this instead of bash grep/find.
```

### search_tool Description (after)
```
PREFERRED over bash grep/find/rg for code search. Search files and file contents
using regex or literal patterns. Returns structured {path, line, content} matches
with line numbers. Honors permission gates and applies per-tool output caps.
Supports output modes: files_with_matches (paths only), content (lines with
numbers), count (match counts).
```

### bash Description (before)
```
Execute a bash command.
```

### bash Description (after)
```
For code investigation (searching source, reading files, listing directories),
prefer the structured `search_tool`, `read_file`, and `glob` tools. Execute a
bash command.
```

### Why
LLMs read tool descriptions at registration time to decide which tool to use. The updated descriptions give the LLM a clear preference signal before it ever makes a call.

---

## Phase 4: Prompt Updates

**What**: Updated both coder and planner prompts with explicit search_tool preference guidance.

**Files changed**: `internal/assets/prompts/instruction-coding.md`, `internal/assets/prompts/instruction-planning.md`

### Coder Prompt (new additions)
```
- **PREFERRED**: Use `search_tool` ... instead of `bash` with `grep`, `find`, or `rg`.
  `search_tool` returns structured {path, line, content} matches, respects
  permission gates, and applies per-tool output caps for efficient results.
- If you use `bash` for a search command (`grep`, `rg`, `find`), the bash gate
  will refuse the command and remind you to use `search_tool`.
```

### Planner Prompt (new additions)
```
* **YOU CAN**: Read files, search the codebase with `search_tool` (preferred
  over bash+grep/find/rg), list directories, and analyze project structure.
* **SEARCH PREFERENCE**: For code search and pattern matching, always use
  `search_tool` instead of `bash` with `grep`, `rg`, or `find`. `search_tool`
  returns structured results and respects permission boundaries. The bash gate
  will block search commands in `bash` and redirect you to `search_tool`.
```

---

## Phase 5: Double-Read Performance Fix

**What**: Content mode previously read every matching file twice (once in `countMatches`, once in `readFileContent`). Now reads once.

**Files changed**: `internal/tool/search.go`

### Before
```
content mode → countMatches(path) [reads file] → readFileContent(path) [reads file AGAIN]
```

### After
```
content mode → readFileContentWithCount(path) [reads file ONCE, counts + formats]
```

A new `readFileContentWithCount` function combines counting and output formatting into a single file read. The old `readFileContent` is kept for backward compatibility.

---

## Validation

```
go build ./...                    ✅ passes
go test ./internal/tool/...       ✅ passes
go test ./internal/tool/ast/...   ✅ passes
go test ./internal/config/...     ✅ passes
go test ./internal/executor/...   ✅ passes
go test ./internal/agent/...      ✅ passes
```

## Files Changed

| File | What changed |
|------|-------------|
| `internal/tool/search.go` | Description, schema (exclude, recursive, output_mode aliases), walkFn refactor, double-read fix, readFileContentWithCount |
| `internal/tool/implementations.go` | Bash gate (isSearchCommand, extractCommand, getBashGateLevel), ValidateBashCommand update, WrapError passthrough, bash description |
| `internal/assets/prompts/instruction-coding.md` | PREFERRED search_tool guidance, bash gate notice |
| `internal/assets/prompts/instruction-planning.md` | Search preference guidance |

mlhher:
I'll try it out. Leave a bit please so I can see where this goes. The original grep tool was extremely simple that is correct. Will need to see how this behaves with Qwen and Gemma specifically.

giveen:
image
search_tool being used by Gemma4-26B

giveen:
image
search_tool being used by Qwen3.6-27B


Looks like Qwen3.6-27B tried to use it beforte just swtiching to other tools as well, like `find`

giveen:
image

second run using Qwen3.6-27B

internal/assets/prompts/instruction-coding.md
```diff
@@ -1,24 +1,31 @@
...
 
 ## Capabilities
+
 - You have access to the same tools as the main agent, **IN ADDITION** you also have access to file-modifying tools (`write_file`, `target_edit`) that are withheld from the main agent.
 - You should use `read_file` to understand the context.
 - You should use `write_file` or `target_edit` to modify code as instructed.
 - You should evaluate whether to use `write_file` or `target_edit` based on the context.
-- You must prefer native tools (e.g. `write_file` and `target_edit`) over bash commands (e.g. `echo` and `sed`).
+- You must prefer native tools (e.g. `search_tool`, `write_file`, and `target_edit`) over bash commands (e.g. `grep`, `find`, `echo`, and `sed`).
+- **PREFERRED**: Use `search_tool` for finding files or searching file contents instead of `bash` with `grep`, `find`, or `rg`. `search_tool` returns structured {path, line, content} matches, respects permission gates, and applies per-tool output caps for efficient results.
```
mlhher: From your comments I see why you added this but I'd still try to leave it out if possible (the above line should be enough, presumably even with a refusal once). Having the model once try grep and then getting refused should not be a deal breaker the model should pivot on its own (and we already basically instruct it to to use the search_tool instead of bash). If the model consistently fails even after getting refused we might need to re-evaluate, I am not a fan of forcing such specific behaviour through prompting.

internal/assets/prompts/instruction-coding.md
```diff
@@ -1,24 +1,31 @@
...
+- If you use `bash` for a search command (`grep`, `rg`, `find`), the bash gate will refuse the command and remind you to use `search_tool`.
```
mlhher: Preferably use something like 

> If you use `bash` for a search command (e.g. `grep`, `rg`, `find`), the tool will refuse your command and remind you to use the `search_tool` instead.

internal/assets/prompts/instruction-planning.md
```diff
@@ -3,49 +3,62 @@ You are the **Lead Architect and Planning Agent**.
...
+
 **CRITICAL: You are an ARCHITECT, not a CODER.**
 
-*   **YOU CAN**: Read files, search the codebase, list directories, and analyze project structure.
-*   **YOU MUST**: Use `write_implementation_plan` to record your design before any execution.
-*   **YOU MUST**: Use `spawn_subagent` (type `coder`) for **ALL** direct file modifications. **CRITICAL TOOL RULE: You MUST invoke the `spawn_subagent` tool MULTIPLE TIMES—exactly once for EVERY individual step in your Implementation Plan. You are strictly forbidden from passing multiple steps or the entire plan into a single `spawn_subagent` call.**
-*   **YOU CANNOT**: Edit files, create files (other than the plan), or run destructive bash commands.
-    *   *Note: Direct file-editing tools (like `write_file` or `target_edit`) are physically removed from your toolset. You MUST delegate all coding to subagents.*
-    *   *Even for requests to "implement", "add", "update", or "edit", you MUST follow the plan -> subagent pipeline. Direct edits are only for subagents.*
+* **YOU CAN**: Read files, search the codebase with `search_tool` (preferred over bash+grep/find/rg), list directories, and analyze project structure.
```
mlhher: Please remove the (preferred over bash+grep/find/rg) here as that is explained in the next block.

internal/assets/prompts/instruction-planning.md
```diff
@@ -3,49 +3,62 @@ You are the **Lead Architect and Planning Agent**.
...
+* **SEARCH PREFERENCE**: For code search and pattern matching, always use `search_tool` instead of `bash` with `grep`, `rg`, or `find`. `search_tool` returns structured results and respects permission boundaries. The bash gate will block search commands in `bash` and redirect you to `search_tool`.
```
mlhher: Preferably use something like:

> **YOU MUST**: Use the search_tool instead of using the bash_tool with e.g. grep/find/rg to search for and match patterns and strings...

or similar, gladly try out what gives the best result often some minor wording can give quite some boosts in consistency.

internal/tool/implementations.go
```diff
@@ -245,7 +311,7 @@ func shellDisplayName() string {
 
 func (t ShellTool) Name() string { return "bash" }
 func (t ShellTool) Description() string {
-	return fmt.Sprintf("Execute a %s command.", shellDisplayName())
+	return fmt.Sprintf("For code investigation (searching source, reading files, listing directories), prefer the structured `search_tool`, `read_file`, and `glob` tools. Execute a %s command.", shellDisplayName())
```
mlhher: This should not be included it in the Bash tools description, it might bias the model. Simply restore the original here. If the model gets refused it should pivot on its own properly.

mlhher:
Looks good I am fine with it. Please try out my proposed changes, I will also try them out once (Qwen3.6-35B-A3B and Gemma4-26B-A4B) and if everything works well this will be merged.

Thank you!

giveen:
> Looks good I am fine with it. Please try out my proposed changes, I will also try them out once (Qwen3.6-35B-A3B and Gemma4-26B-A4B) and if everything works well this will be merged.
> 
> Thank you!

Let me go through your comments and make adjustments.

mlhher:
Thank you again, seems to work great with both Qwen and Gemma

PR #84, opened 2026-06-23
fix: sort skill enum alphabetically in activate_skill tool

schelmo:
#### Problem

The `activate_skill` tool's parameter schema included an `enum` list of available skill names. Because the list was built by iterating over a Go `map[string]*skill.Skill`, the iteration order was randomized on every call. This caused the enum items to "shift" between LLM prompts, confusing the model and resulting in full prompt re-processing when using llama-server

#### Fix

sort the skill names


mlhher:
Great catch thank you!

PR #85, opened 2026-06-23
feat: add skill reference discovery

schelmo:
#### Summary
This commit introduces the ability for skills to carry reference files (e.g. API docs, guides, conventions) alongside their `SKILL.md`, and provides a tool to read those references after a skill is activated.

#### Changes

**New: `skill_read_reference` tool** (`internal/tool/skill_reference_tool.go`)
- Reads reference files from an activated skill's directory
- Includes path traversal protection and output truncation at 32 KB

**New: `DiscoverSkillReferences` function** (`internal/skill/skill.go`)
- Walks a skill directory and discovers all reference files
- Excludes `SKILL.md`, dotfiles/dotdirs, and `scripts/` (scripts are already handled)

**Updated: `ActivateSkillTool`** (`internal/tool/skill_tool.go`)
- Now lists available reference files in its response after activation
- Tells the user how to read them via the new tool

**Updated: Executor** (`internal/executor/executor.go`)
- Registers the new `skill_read_reference` tool

**Tests**
- `TestDiscoverSkillReferences` validates file discovery with nested dirs and exclusion rules
- `TestSkillReadReferenceTool` covers happy path, missing skill, missing file, path traversal rejection, and output truncation

#### Why
Skills often need to reference additional documentation (API specs, conventions, etc.). This gives agents first-class access to those files without requiring hardcoded paths or manual file reads.

### DIFF
--- a/internal/skill/skill.go
+++ b/internal/skill/skill.go
@@ -131,7 +131,7 @@ func DiscoverSkillReferences(s *Skill) []string {
 			return filepath.SkipDir
 		}
 		// Skip SKILL.md itself
-		if filepath.Base(path) == "SKILL.md" {
+		if base == "SKILL.md" {
 			return nil
 		}
 		// Only include files, not directories
--- a/internal/skill/skill.go
+++ b/internal/skill/skill.go
@@ -110,6 +110,44 @@ func parseSkillFile(content string) (*SkillMetadata, string, error) {
 	return &metadata, strings.TrimSpace(body.String()), nil
 }
 
+// DiscoverSkillReferences returns all files in the skill directory
+// (excluding SKILL.md, dotfiles, dotdirs, and scripts/).
+func DiscoverSkillReferences(s *Skill) []string {
+	var files []string
+	_ = filepath.Walk(s.Path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return nil
+		}
+		base := filepath.Base(path)
+		// Skip dotfiles and dotdirs
+		if strings.HasPrefix(base, ".") {
+			if info.IsDir() {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		// Skip scripts/ directory — handled separately as executable tools
+		if base == "scripts" {
+			return filepath.SkipDir
+		}
+		// Skip SKILL.md itself
+		if filepath.Base(path) == "SKILL.md" {
+			return nil
+		}
+		// Only include files, not directories
+		if info.IsDir() {
+			return nil
+		}
+		rel, err := filepath.Rel(s.Path, path)
+		if err != nil {
+			return nil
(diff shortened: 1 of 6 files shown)


mlhher:
Thank you for your contribution. Can you explain the change further? I am fine with adding it I just want to confirm my understanding. Is this documented in https://agentskills.io/ somewhere, and how do other harnesses do this? Is this generally expected of a harness?

schelmo:
yeah it is documented here: https://agentskills.io/specification#references/ (assets below it)

i thought having the whole file tree (except SKILL.md, scripts, dotfiles and -dirs) is better. i had an implementation before which just lists references/assets/etc. which are mentioned in the SKILL.md, but that got out of hand parsing the markdown.

for skills its especially useful to keep the SKILL.md short to let the agent have a brief overview and if it is interested more, it can get more information from references (like API documentation, deeper documentation) or assets like for example branding files, like svgs, etc.

references are widely used in existing skill libraries/repositories

mlhher:
Thank you! Let me try it out a bit and evaluate the code, if everything seems good this will be merged.

internal/skill/skill_test.go
```diff
@@ -62,6 +63,38 @@ Instructions here.
...
+	os.WriteFile(filepath.Join(tmpDir, "m.md"), []byte("m"), 0644)
+
+	skill, err := LoadSkill(tmpDir)
+	if err != nil {
+		t.Fatalf("LoadSkill failed: %v", err)
+	}
+
+	refs := DiscoverSkillReferences(skill)
+
+	expected := map[string]bool{"a-dir/a.md": true, "m.md": true, "z-dir/z.md": true}
```
mlhher: This will presumably fail on Windows due to the forward slash. Instead use filepath.Join please.

internal/skill/skill.go
```diff
@@ -110,6 +110,44 @@ func parseSkillFile(content string) (*SkillMetadata, string, error) {
...
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		// Skip scripts/ directory — handled separately as executable tools
+		if base == "scripts" {
+			return filepath.SkipDir
+		}
+		// Skip SKILL.md itself
+		if filepath.Base(path) == "SKILL.md" {
```
mlhher: `base` should still be equal to `filepath.Base(path)` here so you should be able to avoid the extra call here and just use `base == "SKILL.md"`

mlhher:
@schelmo Looks good. One change needed for Windows and one minor note. Will be merged afterwards. Thank you again!

schelmo:
both changes are done. thank you

schelmo:
wait found another bug in the implementation_test.go

schelmo:
okey fixed a missing import

PR #88, opened 2026-06-26
feat: add .gitignore support to search_tool for local LLM efficiency

giveen:
Add a self-contained .gitignore parser and matcher (Go stdlib only) to search_tool so it automatically skips build artifacts, vendored deps, and generated output when searching inside a git repo.

Why this matters for local LLMs with limited context windows:
- Every result competes for precious context tokens. Without gitignore, a search on a project with dist/ and vendor/ dirs floods the LLM with noise before it sees real code. With gitignore, only relevant files.
- Files in ignored dirs are never read or even walked into, so the tool spends zero time (and zero output tokens) on them.
- Fewer results means fewer round-trips: the LLM gets the answer from one search call instead of having to refine.

Benchmark results (Intel Core Ultra 9 285K):
  Output reduction:  93% — 220 files (14,170 bytes, ~3,542 tokens) dropped to 20 files (1,090 bytes, ~272 tokens) on a project with 20 source + 200 generated/vendor files, all matching the search pattern Repo root caching: 7,209x faster — cold walk 4,992ns vs cached 0.69ns (0 allocations vs 23), so repeated search_tool calls in a session cost essentially nothing for the cache
  Schema size:        346 fewer chars (~86 tokens) saved by removing the
                      no_gitignore param and tightening descriptions

Implementation:
- gitignore.go: parses .gitignore patterns, compiles them to Go regex, supports: *, ?, **, [...], negation (!), dir-only (/), anchored patterns (/prefix), comments, backslash escaping
- sync.Once caches the repo root at process startup — no redundant directory walking across multiple search_tool calls
- getGitIgnoreForPath uses cached root when search path is inside it, falls back to per-path resolution for external paths
- matchesGitIgnore convenience wrapper: nil-safe, single filepath.Rel
- Removed struct-field caching (SearchTool is a singleton, no stale state risk)
- Removed no_gitignore parameter (saves ~86 tokens from JSON schema; gitignore is silently correct by default)
- Tightened descriptions across tool schema to save LLM context

Tests: 21 new tests covering pattern semantics, file loading, repo root discovery, full integration with search_tool, and a benchmark suite (TestGitIgnoreOutputSavings, BenchmarkSearchOutputSize, BenchmarkFindRepoRoot, BenchmarkSchemaSize) for ongoing measurement.

- [ X] **By checking this box, I confirm that I have read and agree to the terms of the [CLA.md](https://github.com/mlhher/late/blob/main/.github/CLA.md) in this repository.** *(To check the box, put an `x` between the brackets like this: `[x]`)*

### DIFF
--- a/.llmignore
+++ b/.llmignore
@@ -1,0 +1,38 @@
+# .llmignore - Generated at rival.tips/llmignore
+# Files and patterns AI agents should never read or send to LLMs.
+# Syntax: identical to .gitignore (fnmatch patterns)
+
+# Secrets & credentials
+.env
+.env.*
+**/*.pem
+**/*.key
+**/credentials.*
+**/secrets.*
+**/.secret*
+
+# Sensitive data
+**/migrations/*.sql
+**/fixtures/**
+**/seed-data/**
+
+# Large binaries
+**/*.wasm
+**/*.bin
+**/*.sqlite
+**/*.db
+
+# Dependency Lockfiles
+**/go.sum
+
+# Vendor directories
+**/vendor/**
+
+# Large test fixtures
+testdata/
+*_test.json
+*.golden
+
+# Pre-compiled or generated assets
+mocks/
+*_mock.go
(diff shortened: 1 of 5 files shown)


mlhher:
Will look over it thanks!

PR #89, opened 2026-06-29
add a option to skip adding system prompt

rhjdvsgsgks:
### Description of Changes

it is useful when i just want a chat bot rather than a coding agent


mlhher:
I sometimes do something similar myself by using e.g. -system-prompt "You are a helpful assistant." Is this basically just a convenience switch?

mlhher:
@rhjdvsgsgks 

While I do get the reasoning, I am not quite sure as this feels like simply duplicating the -system-prompt flag. If there is something I am missing feel free to re-open this and explain.

PR #98, opened 2026-07-13
fix(mcp): resolve all MCP spec compliance issues

giveen:
### Description of Changes

- Replace manual stdio transport with SDK CommandTransport for proper SIGTERM → wait → SIGKILL process lifecycle
- Reuse single SDK client instance across all server sessions
- Wire ToolListChangedHandler to re-discover tools on dynamic changes
- Handle all MCP content types (AudioContent, ResourceLink, EmbeddedResource) plus default warning for unknown types
- Add SSE and Streamable HTTP transport support for remote servers
- Capture stderr from stdio subprocesses on connection failure
- Log warning on config file write failure instead of silent discard
- Suppress unnecessary 'roots' capability advertisement

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -10,6 +10,8 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"late/internal/common"
@@ -17,6 +19,7 @@ import (
 
 // Client manages MCP connections and tools.
 type Client struct {
+	mu        sync.RWMutex
 	sessions  map[string]*mcp.ClientSession
 	tools     map[string]*ToolAdapter
 	sdkClient *mcp.Client // single SDK client instance managing all sessions
@@ -148,21 +151,27 @@ func (c *Client) Connect(ctx context.Context, transport mcp.Transport, serverNam
 		return fmt.Errorf("failed to connect to MCP server: %w", err)
 	}
 
-	// Store session keyed by server name so Close() shuts down every
-	// subprocess, not just the last one connected.
-	c.sessions[serverName] = session
-
-	// List and store tools using iterator
+	// Collect adapters without holding the lock; the SDK's Tools iterator may
+	// perform RPCs, and we want to avoid blocking readers.
+	var adapters []*ToolAdapter
 	for tool := range session.Tools(ctx, &mcp.ListToolsParams{}) {
 		if tool != nil {
-			adapter := &ToolAdapter{
+			adapters = append(adapters, &ToolAdapter{
 				mcpTool:    tool,
 				session:    session,
 				serverName: serverName,
-			}
-			c.tools[adapter.Name()] = adapter
+			})
 		}
 	}
+
+	// Store session keyed by server name so Close() shuts down every
+	// subprocess, not just the last one connected.
+	c.mu.Lock()
+	c.sessions[serverName] = session
+	for _, adapter := range adapters {
+		c.tools[adapter.Name()] = adapter
+	}
+	c.mu.Unlock()
 
 	return nil
 }
@@ -173,6 +182,7 @@ func (c *Client) handleToolListChanged(ctx context.Context, req *mcp.ToolListCha
 	sess := req.Session
 
 	// Map the SDK session back to the server name we assigned in Connect.
+	c.mu.RLock()
 	var serverName string
 	for name, s := range c.sessions {
 		if s == sess {
@@ -180,6 +190,7 @@ func (c *Client) handleToolListChanged(ctx context.Context, req *mcp.ToolListCha
 			break
 		}
 	}
+	c.mu.RUnlock()
 	if serverName == "" {
 		fmt.Fprintf(os.Stderr, "MCP tool list changed notification for unknown session\n")
 		return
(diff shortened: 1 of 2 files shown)


mlhher:
@giveen 

Great one thanks!

 I evaluated the code a bit and noticed two things that might cause issues, please take a look at them:

 - handleToolListChanged modifies the c.tools map asynchronously, so we need a sync.RWMutex on the Client struct to prevent panics.
 - stderrBuf races because a background routine writes to it while ConnectFromConfig reads it on failure, so we should wrap the buffer with a mutex.

Also a minor thing but would be good to have a tiny time.Sleep before checking the buffer, since an instant failure might read it before the bg routine even writes the error.

Will try this out a little more afterwards and if nothing new pops up it will get merged, thanks!

giveen:
@mlhher

internal/mcp/client.go
```diff
@@ -81,10 +101,19 @@ func (t *ToolAdapter) Execute(ctx context.Context, args json.RawMessage) (string
...
+		case *mcp.ImageContent:
+			sb.WriteString(fmt.Sprintf("[Image: %s]", c.MIMEType))
+		case *mcp.AudioContent:
+			sb.WriteString(fmt.Sprintf("[Audio: %s]", c.MIMEType))
+		case *mcp.ResourceLink:
+			sb.WriteString(fmt.Sprintf("[Resource: %s]", c.URI))
+		case *mcp.EmbeddedResource:
+			sb.WriteString("[Embedded resource]")
+		default:
+			fmt.Fprintf(os.Stderr, "MCP tool returned unhandled content type: %T\n", content)
```
mlhher: This will bleed into the TUI. Please remove them or add them to the TUI but do not use raw printers (stdout/stderr etc.). Note the comments below too. Might even be fair to just comment these out for now (for the ones not immediately necessary for the user to see) so we can eventually replace them when we have a global logger.

internal/mcp/client.go
```diff
@@ -112,42 +141,93 @@ func (t *ToolAdapter) CallString(args json.RawMessage) string {
...
+	var serverName string
+	for name, s := range c.sessions {
+		if s == sess {
+			serverName = name
+			break
+		}
+	}
+	c.mu.RUnlock()
+	if serverName == "" {
+		fmt.Fprintf(os.Stderr, "MCP tool list changed notification for unknown session\n")
```
mlhher: Similarly this would also bleed into the TUI

internal/mcp/client.go
```diff
@@ -112,42 +141,93 @@ func (t *ToolAdapter) CallString(args json.RawMessage) string {
...
+		return
+	}
+
+	fmt.Printf("MCP server '%s' tools changed, re-discovering...\n", serverName)
```
mlhher: And this as well. Note that you can just remove this one (it should just happen no need to notify the user at all).

internal/mcp/client.go
```diff
@@ -166,53 +248,126 @@ func (c *Client) GetTool(name string) tool.Tool {
...
 
-// NewStdioTransport creates a new transport that communicates with a subprocess.
+// lockedBuffer wraps a bytes.Buffer with a mutex so it can be safely written
+// by a background goroutine and read by another goroutine on connection failure.
+type lockedBuffer struct {
+	mu  sync.Mutex
+	buf bytes.Buffer
+}
+
+func (b *lockedBuffer) Write(p []byte) (int, error) {
```
mlhher, line 273: This should be capped otherwise it can grow forever. Something like this:

```go
        const maxSize = 8192
        if b.buf.Len() < maxSize {
            writeLen := len(p)
            if b.buf.Len()+writeLen > maxSize {
                writeLen = maxSize - b.buf.Len()
            }
            b.buf.Write(p[:writeLen])
        }
        
        // returning len(p) so the pipe keeps draining
        return len(p), nil
```

internal/mcp/config.go
```diff
@@ -43,8 +47,9 @@ func LoadMCPConfig() (*MCPConfig, error) {
 		defaultData, _ := json.MarshalIndent(emptyConfig, "", "  ")
 
 		if err := os.MkdirAll(lateConfigDir, 0700); err == nil {
-			// Ignore write error, just fallback to empty config
-			_ = os.WriteFile(defaultUserPath, defaultData, 0600)
+			if err := os.WriteFile(defaultUserPath, defaultData, 0600); err != nil {
+				fmt.Fprintf(os.Stderr, "Warning: Could not write default MCP config to %s: %v\n", defaultUserPath, err)
```
mlhher: Again this one too. Depending on how noticeable these issues are to the user (e.g. mcp server is not connected) they potentially should be notified by a tiny message. In other cases they might not need to be informed.

I will add logging later on we clearly do need it. But currently we do not have the appropriate infrastructure (a global logger).

mlhher:
Might have missed these previously. But I think this is all now. Thank you a lot!

internal/mcp/client.go
giveen: Resolved: commented out this raw stderr print. Will migrate to a global logger when infrastructure exists.

internal/mcp/client.go
giveen: Resolved: commented out — this should just happen silently.

internal/mcp/client.go
giveen: Resolved: capped lockedBuffer.Write at 8192 bytes as suggested. Pipe keeps draining by returning len(p).

internal/mcp/client.go
giveen: Resolved: commented out this raw stderr print.

internal/mcp/config.go
giveen: Resolved: commented out. Will route through global logger when available.

mlhher:
👍

PR #105, opened 2026-07-27
feat(plugin): universal plugin system — OMP and Claude Code compatibility

giveen:
### What

The Late plugin system now supports installing plugins from **three ecosystems** — not just plugins written specifically for Late:

| Format | Manifest | Auto-detected surfaces |
|--------|----------|----------------------|
| **Late** (native) | `package.json` → `"late"` field | skills, commands, MCP, hooks, themes, tools |
| **OMP** (Oh My Pi) | `package.json` → `"omp"` field | skills, commands, MCP, hooks |
| **Claude Code** | `.claude-plugin/plugin.json` | `skills/`, `commands/`, `.mcp.json` (flat + wrapped), `hooks/hooks.json` |

### Why

Late should be able to install plugins from any ecosystem without requiring authors to publish specifically for Late. This makes the entire OMP and Claude Code plugin catalogs available to Late users.

### How

- `internal/plugin/manifest.go` — Added `OmpManifest` and `ClaudePluginManifest` structs. `LoadPlugin` now tries three format loaders in order: native Late → OMP → Claude Code. Both OMP and Claude Code formats are translated into `LateManifest` so all surface registration (skills, MCP, commands, hooks) works identically.
- `.mcp.json` detection handles both `{"mcpServers": {...}}` (wrapped) and `{"name": {...}}` (flat Claude Code convention) formats.
- `internal/plugin/installer.go` — Fixed relative symlink path for scoped npm packages (`@scope/name`).
- `internal/plugin/manager.go` — `discoverFromDir` now accepts symlinks (Go `os.ReadDir().IsDir()` returns false for symlinks) and recurses into `@`-prefixed scope directories.
- `internal/plugin/hooks.go` — Hook subprocesses get sandboxing via `setCmdSysProcAttr`.
- `internal/plugin/sandbox_linux.go` — Go 1.26 compat: `NoNewPrivs` removed from `syscall.SysProcAttr`.

### Verified With Real Plugins

**OMP plugin from npm:**
```
late plugin install @a5c-ai/babysitter-omp
→ 1 skill(s) detected
```

**Claude Code official plugins:**
```
late plugin install ./asana     (.claude-plugin/plugin.json + flat .mcp.json)
→ 1 MCP server(s) detected

late plugin install ./discord   (.claude-plugin/plugin.json + skills/ + wrapped .mcp.json)
→ 1 skill(s), 1 MCP server(s) detected
```

All standard plugin lifecycle commands verified: `install`, `link`, `list`, `disable`, `enable`, `remove`, `update`.

### Open Items

- Claude Code `hooks/hooks.json` auto-detection is stubbed but not yet wired into the hook execution pipeline (the hook event model differs between systems).
- No marketplace registry published yet (the default endpoint `registry.late.dev` is a placeholder that falls through to npm).

### CLA


giveen:
@mlhher we had talked about a plugin system, so here you go, its compatible with OMP and Claude Code plugins plus gives us room to make our own plugins as well.

mlhher:
Lol @giveen what a champ. This will take some time to go over everything and test everything so please bear with me.

giveen:
No worries, I was on vacation and this was my work every morning with my coffee at the beach. 

I've only done preliminary tests with it, but more in depth testing should be done. 

Im going to be installing and uninstall lots of plugins this week to see what screw ups I made.

giveen:
@mlhher  okay I "think" thats everything.

mlhher:
@giveen Due to the size I will go at this in multiple passes. For now these are things that I did notice. Please take a look at them.

- **`LoadPlugin` → `InstallFromLocal` / `removeFromDir`:** A manifest's plugin name is only checked for emptiness, then used in `filepath.Join`. Names containing `..` can make install/remove operate outside the plugin store.
- **`isPluginCmd` → `CommandHandler`:** The full input is matched before arguments are split. For example, `/lint file.go` does not match the registered `/lint`, so handlers cannot receive arguments.
- **`pluginInlineTool.RequiresConfirmation`:** This always returns `false`, allowing arbitrary plugin scripts to run without the normal tool confirmation. The documentation notes it should ask for permission. Please clarify which one is correct and settle them on whatever behavior is expected.
- **`PluginManager.HandleCommand` → `PluginManager.All`:** `HandleCommand` holds `RLock` and then calls `All()`, which acquires it again. A queued writer can deadlock this nested read lock.
- **`runHook`:** All hook and inline-tool stdin payloads are limited to 256 bytes. Normal messages, tool calls, results, and arguments can easily exceed this.
- **`CallOnToolResultHooks`:** Plain-text tool results are wrapped in `json.RawMessage`. `json.Marshal` then fails, and its error is ignored, so the hook receives an empty payload.
- **`CallOnInputHooks` / `CallOnTurnStartHooks` / `CallOnTurnEndHooks`:** These hooks are defined and documented but are not called by production code. For now I'd suggest removing them entirely as they can be added later on. If you want to add them now please wire them up so plugins can use them.
- **`TestInstallFromGit_CleansUpOnCloneFailure`:** This test clones a nonexistent GitHub repository and can prompt for credentials during `make test`. Please replace the live network call with a local fake Git command or repository so the test is deterministic and non-interactive.

mlhher:
👀

mlhher:
@giveen Please mind this taking longer, it needs to be checked fully and I am currently away and quite busy.

giveen:
> @giveen Please mind this taking longer, it needs to be checked fully and I am currently away and quite busy.

no worries, its not going anywhere :)

mlhher:
@giveen 

1. Inline plugin tools are registered as `plugin:tool`, but OpenAI-compatible endpoints reject names containing `:`. Sanitize names, enforce length limits, and handle collisions. Please also check how the MCP tool handles it as there was the same issue reported by a user and afterwards presumably fixed. `internal/plugin/tools.go:56`, `internal/session/session.go:76`

2. Adding plugin MCP servers reconnects existing MCP servers a second time. The original sessions are overwritten without being closed, leaking their processes. `cmd/late/main.go:267`, `cmd/late/main.go:315`, `internal/mcp/client.go:182`

3. Git plugin updates delete the working version before moving and validating the replacement. A broken update therefore destroys the working installation without rollback. `internal/plugin/installer.go:412`, `internal/plugin/installer.go:415`, `internal/plugin/installer.go:419`

4. Reloading, disabling, or removing a plugin only refreshes commands and themes. Its tools, MCP sessions, hooks, and skills remain active or registered. `cmd/late/main.go:555`, `cmd/late/main.go:568`

5. `onMessageSend` hooks never run because message sending does not call `ApplyMessageHook`. `internal/tui/model.go:212`, `internal/tui/update.go:1126`

6. Plugin skill cleanup deletes every unrecognized symlink in Late’s skills directory, including unrelated symlinks created by the user. `internal/plugin/manager.go:320`, `internal/plugin/manager.go:331`

7. Tool-call confirmation happens before `onToolCall` hooks mutate the arguments. A plugin can therefore change the arguments after the user approved them. `cmd/late/main.go:584`, `internal/plugin/hooks.go:227`

8. `--project` silently installs globally when the project’s `.late/plugins` directory does not already exist. `internal/plugin/manager.go:53`, `cmd/late/main.go:165`

9. Git plugins are stored using the repository name but removed using the manifest name. Removal fails when those names differ. `internal/plugin/installer.go:99`, `internal/plugin/installer.go:528`

10. Ignore per discussion: after merging main, help advertises `/clear` while the handler still implements `/new`. `internal/tui/state.go:61`, `internal/tui/update.go:902`

11. `late plugin install` and `late plugin link` panic when called without an argument because `parseProjectFlag` reads `args[0]`. `internal/plugin/command.go:183`, `internal/plugin/command.go:255`, `internal/plugin/command.go:304`

12. Registering the first plugin skill fails when Late’s skills directory does not already exist because the directory is never created. `internal/plugin/manager.go:219`, `internal/plugin/manager.go:311`

13. Skills from scoped plugins produce nested symlink paths, but the required parent directories are not created, so registration fails. `internal/plugin/manager.go:303`, `internal/plugin/manager.go:306`

14. Linking a local plugin writes Late’s metadata into the developer’s source directory through the symlink. That cached metadata then overrides later manifest changes, including after restarting Late. `internal/plugin/installer.go:191`, `internal/plugin/manifest.go:545`

15. When Late starts with zero plugins, it creates no watcher. Installing the first plugin while Late is running is therefore not detected until restart. `cmd/late/main.go:275`, `cmd/late/main.go:546`

16. Plugin command handlers run inside the TUI update loop. A slow command freezes input and rendering until the command finishes or times out. `internal/tui/update.go:1050`, `internal/plugin/hooks.go:21`

17. Claude Code `hooks/hooks.json` is parsed and then discarded. The docs say it is supported; remove this integration fully for now so it can later be implemented cleanly on top. `internal/plugin/manifest.go:476`, `internal/plugin/manifest.go:482`

18. Claude Code command Markdown files contribute command names, but their contents are never used. Preferably remove the entire Claude Code integration so it can later be implemented cleanly on top. `internal/plugin/manifest.go:434`, `internal/plugin/manifest.go:441`

19. The documentation describes `onToolResult` as observation-only, result-mutating, and not yet integrated in different places. The implementation does integrate it and allows it to replace results. `docs/plugin-example.md:274`, `docs/plugin-sdk.md:557`, `docs/plugin-sdk.md:631`, `internal/plugin/hooks.go:260`

20. One section documents `onSessionStart` as receiving session metadata, but the implementation sends empty stdin. `docs/plugin-sdk.md:220`, `internal/plugin/hooks.go:256`

21. `runHook` passes the resolved script path twice to `exec.CommandContext`, causing the script to receive an unexpected positional argument in `$1`. `internal/plugin/hooks.go:76`

I hope with these done we might have the vast majority but due to the size I would not necessarily bet on it.

mlhher:
Fully resolved: 2, 3, 5, 6, 7, 8, 9, 11, 14, 16, 17, 19, 20, 21.

  Partially resolved:

  - 1: Names are sanitized and limited, but collisions are handled separately for MCP and inline tools; cross-source
    collisions can still overwrite tools. Existing server:tool disable settings also no longer match renamed
    server__tool tools. cmd/late/main.go:411, cmd/late/main.go:744

  - 4: Tools and MCP servers refresh on removal, but onToolCall hooks are snapshotted once and remain stale. Changed
    MCP configuration also does not reconnect an existing same-named server. internal/plugin/hooks.go:207, internal/
    mcp/client.go:511

  - 12–13: The directories and scoped symlink parents are now created, but the skill scanner ignores symlinked skills
    and does not recurse through scoped parents, so plugin skills still do not reach the agent. internal/skill/
    skill.go:168

  - 15: The zero-plugin watcher now starts, but it still ignores npm/local plugin symlinks; only directory-based
    installs such as Git are detected. internal/plugin/watcher.go:159

  - 18: Claude command Markdown handling was removed, but Claude plugin detection, skills, and MCP integration remain.
    internal/plugin/manifest.go:306

Please also investigate these:

  1. Scoped npm updates use the wrong directory. Updating @scope/plugin installs into plugins/@scope/node_modules/...
     and creates plugins/@scope/@scope/plugin instead of updating plugins/@scope/plugin. internal/plugin/
     installer.go:370, internal/plugin/installer.go:456

  2. onMessageSend can freeze the TUI. Every outgoing message runs these hooks synchronously; each hook can block the
     interface for 15 seconds. internal/tui/update.go:1388, internal/plugin/hooks.go:349

  3. Hook output is unlimited in memory. stdout and stderr are fully buffered before stderr is truncated, so a noisy or
     malicious script can exhaust Late’s memory. internal/plugin/hooks.go:87

  4. Dynamic MCP tool changes do not reach the agent registry. MCP updates its internal tool map after tools/
     list_changed, but the session registry only refreshes after a plugin filesystem change. internal/mcp/
     client.go:237, cmd/late/main.go:601

giveen:
@mlhher  sorry been really busy, I hope to hit this up later this week or maybe next week.

mlhher:
Thanks for the update and no worries, there's no need to rush. I'd rather have something well made that takes longer than something made in a hurry that has issues.

giveen:
@mlhher pushed fixes for everything from your last pass.

Carryover items:
- **1 (tool naming):** inline tools now dedupe against already-registered MCP tool names too, not just other inline tools — a plugin can no longer silently clobber an MCP tool with the same namespaced name. Also added a fallback so old `server:tool` disable-config keys still match the new `server__tool` names.
- **4 (stale hooks / MCP reconnect):** onToolCall/onToolResult middlewares now rebuild on every plugin change instead of only at startup, so hook edits take effect without a restart. `mcp.Client.Reconcile` now detects when a same-named server's command/args/env/url/dir changed and reconnects it instead of leaving it running on the old config.
- **12-13 (plugin skills not reaching the agent):** found the actual bug — `DiscoverSkills` was gating on `entry.IsDir()`, which Go reports as `false` for a symlink even when it points at a directory, so every plugin skill (which are always registered as symlinks) was silently skipped. It also wasn't recursing into scoped plugins' extra nesting level. Fixed both, plus a second bug that exposed: skill name validation was checking the plugin-namespaced symlink name instead of the resolved target's real name, which would've rejected every plugin skill anyway even once discovery found them.
- **15 (watcher ignoring npm/local plugins):** same symlink-detection bug, fixed in the watcher's snapshot too — npm/local installs are always symlinks, only git installs are real directories, so only git was ever picked up before.
- **18 (Claude Code removal):** fully removed — detection, skills, MCP surfacing, all gone. Native-Late and omp are untouched.

New items:
1. **Scoped npm updates:** `late plugin update` on a scoped package (`@scope/name`) was computing the wrong root directory and creating a doubly-nested `plugins/@scope/@scope/name` instead of updating the real plugin. Fixed.
2. **onMessageSend freezing the TUI:** moved hook execution off the Bubble Tea update loop (same async pattern already used for plugin slash-command dispatch) — a slow hook script no longer blocks keystrokes/rendering for its duration.
3. **Unbounded hook output:** stdout/stderr are now captured through a bounded writer during the copy from the child process instead of being fully buffered and truncated after the fact.
4. **Dynamic MCP tool changes not reaching the registry:** added an `OnToolsChanged` hook on the MCP client so a live server's own `tools/list_changed` notification propagates to the session registry, instead of only being picked up incidentally by an unrelated plugin filesystem change.

Added regression tests for the symlink/scoped-nesting bugs, the scoped npm update, the legacy tool-enable key fallback, and the bounded hook output writer. Full `go build`/`go vet`/`go test ./...` passes.

mlhher:
@giveen 

Great work!

Basically everything seems to already work already, bar one issue, so the following are essentially just finishing touches.

When they are done I will start doing a couple of rounds of stress testing to be sure things are really fine. If you find the time please also try it out as multiple eyes are always better to catch issues.

What I checked:

1. CLI linking & lifecycle (link, list, remove): Passed
2. Custom slash command dispatch (/greet async execution & toast): Passed
3. Themes (/themes): Found bug (GetRenderer hardcoding LateTheme) and noted unwired palette
4. Inline tools (late.tools + TUI confirmation): Passed
5. Tool call veto hook (onToolCall interceptor): Passed
6. Skills discovery & symlinking (late.skills): Passed
7. Message transform hook (onMessageSend): Passed (with an stderr leak)

Theme switching via /themes currently doesn't take effect. While ApplyTheme properly sets m.Renderer, markdown rendering calls m.GetRenderer(innerWidth) in internal/tui/view.go, which is hardcoded to build from LateTheme (internal/tui/model.go:132). Having GetRenderer build from the active theme's merged style bytes instead of default LateTheme should fix the markdown styling. Also it seems like the palette field in ThemeFile / theme.go isn't actually wired up to lipgloss styles yet (the TUI chrome colors remain hardcoded). 
This is something I'd really like to have but I assume this introduces significant complexity (especially for the full TUI chrome) which is why I'd suggest to remove it entirely for now (at least the user facing parts). That way we can cleanly built on top of it.
 
In internal/plugin/hooks.go:402, HookedMessage prints [%s/onMessageSend/%s] transformed message\n directly to os.Stderr. This leaks into the terminal. Please silence the log, even better if you just make it a no-op (or just comment it) so we can wire it in when we have the logger.

Another thing I noticed, regarding the earlier item 14 (symlink metadata): 
Skipping SavePluginMeta for plugin.SourceType == "local" (manifest.go:435) prevented polluting the linked source tree, but now late plugin disable has no effect on local plugins because the disabled state is never saved to disk. An approach I'd see here is simply adding something like ~/.config/late/plugins.json with a disabled array. That way enabling/disabling works for all kinds of plugins and we could potentially easily extend it if needed without much friction.

And one final thing: Please for now remove the hardcoded registry URL. You can just leave it empty so we can evaluate how to proceed here in a later iteration.

If I missed anything please tell me so I can try more. Thanks again for the great work, looking forward to seeing this merged!

giveen:
@mlhher pushed fixes for all of these.

1. **Theme switching not affecting markdown**: found it — `GetRenderer` (used by `renderMarkdownBlock`, the actual per-block markdown render path) was hardcoded to always build from the bundled `LateTheme`, completely ignoring whatever `ApplyTheme` had installed. It's a separate code path from `m.Renderer` (used only for the fixed-width chat viewport), which is why that one looked correct while block rendering silently stayed on the default theme. Now stores the applied theme's style bytes and `GetRenderer` builds from those instead.

2. **Palette**: agreed, pulled it out entirely rather than half-wire it — removed the `palette` field from plugin theme files, `ThemeInfo`/`ThemeEntry`, and the `_late_palette` staging in `ResolveRenderTheme` (which is what did nothing, since no lipgloss code ever read it back out). Glamour style overrides (the part that actually works) are untouched. Happy to revisit chrome theming as its own thing later.

3. **onMessageSend stderr leak**: silenced — removed the per-hook "transformed message" print. Left a comment for wiring it into a real logger later.

4. **Local plugin disable not persisting**: went with basically your suggestion — added `~/.config/late/plugins.json` holding just a `disabled` name array. `SavePluginMeta` now writes to that instead of no-op'ing for local plugins, `LoadPluginMeta` applies it back on discovery, and `RemovePlugin` clears the entry on removal so a later relink doesn't inherit a stale disabled state. Git/npm plugins are untouched — they still use their existing per-directory `.late-plugin.json`.

5. **Registry URL**: zeroed out `DefaultRegistryBaseURL`. Bare-name installs fall straight through to the npm fallback now unless `LATE_PLUGIN_REGISTRY` is set.

Added regression tests for all of the above (renderer-reflects-theme, disable-survives-reload, disable-cleared-on-remove, registry-defaults-empty). Full `go build`/`go vet`/`go test ./...` passes.

mlhher:
I will test this a bit and if things seem well it will be merged. I did not find any other obvious issue so far. The merge conflicts are minimal so you can either leave them and I will fix them or you can fix them yourself.

Will either report back when issues come up during testing or merge it. Again big thanks!

mlhher:
@giveen I have found some non-obvious issues, some edge cases. I am currently resolving the merge conflict and investigating these. I will push a fix for them when my rounds of local testing are done and then do some more sanity checks. One thing I did remove for now is the hot reloader. Again, similar to the TUI chrome this is also something I am in favour of though I'd rather have it implemented in a separate run so as to ensure this branch lands without issues.

mlhher:
@giveen Thank you again for the great work!

Since you already noted it, please wait with the TUI chrome surface as I am about to push some bigger changes there. I will resolve my conflicts, do some testing and then push. Generally, editing other things should be fine for the most part but I'd still suggest if you want to do something non-trivial please open an issue/discussion first so we can just quickly verify we don't accidentally touch similar ends (at least not in a meaningful way). Further I will publish a registry for now on GH as it takes away a lot of complexity and allows fast iterating. With your implementation that is trivial to change so that is likely the best way forward for now.

giveen:
This may end up being my last push for a while. I've been focused on my own harness focused around ninfer inference engine.

PR #107, opened 2026-07-29
feat(client): add video_url content part for multimodal video input

octo-patch:
Reason: Add a video content-part type and request serialization so the OpenAI-compatible chat schema can express video input, preserving existing text/image handling.

Changes:
- Add `ContentPartVideoURL` content-part type and a `VideoURL` payload struct (mirroring `ImageURL`) in `internal/client/types.go`, serialized as the OpenAI-compatible `video_url` field.
- Wire video attachments into the orchestrator's `Submit` path alongside images: files detected as `video/*` are encoded as `video_url` content parts when the model supports vision; otherwise they fall back to a text note (parallel to the existing image behavior).
- Allow the TUI file picker to attach video files using the same vision-capability gate as images, and keep media attachments consistent on model changes.

Checks:
- `go build ./...`
- `go vet ./internal/client/ ./internal/orchestrator/ ./internal/tui/`
- `go test ./internal/client/ ./internal/orchestrator/ ./internal/tui/`

### DIFF
--- a/internal/client/types.go
+++ b/internal/client/types.go
@@ -95,16 +95,28 @@ type ContentPartType string
 const (
 	ContentPartText     ContentPartType = "text"
 	ContentPartImageURL ContentPartType = "image_url"
+	ContentPartVideoURL ContentPartType = "video_url"
 )
 
 type ContentPart struct {
 	Type         ContentPartType `json:"type"`
 	Text         string          `json:"text,omitempty"`
 	ImageURL     *ImageURL       `json:"image_url,omitempty"`
+	VideoURL     *VideoURL       `json:"video_url,omitempty"`
 	IsAttachment bool            `json:"-"` // Purely for UI filtering
 }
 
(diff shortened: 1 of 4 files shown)


mlhher:
Thank you for your contribution. I see that you removed the PR template. 
As thus I must ask you to manually confirm the CLA.

Will be tested and if nothing comes up can be merged afterwards. Thank you!

giveen:
I'm confused on what an IDE needs with video input.

mlhher:
There are some code -> video workflows where this can come in handy, also generally I think it can be helpful to attach videos as workflows instead of manually explaining things. It is a minimal targeted change and should equally support Gemma, I am not against it.

giveen:
Okay  cool, I was just really curious on how this would work

mlhher:
@octo-patch Just a reminder, this sadly cannot be merged without the CLA.

mlhher:
@octo-patch Cannot be merged due to missing CLA.

PR #110, opened 2026-08-16
Feat/subagent history

### LINKED ISSUE #109: Redesigning session history to set the stage for parallel subagents
Hey all 👋 — I've got a working prototype of this running locally and I'm still in the testing/refining phase. Before I start opening PRs, I wanted to put the shape of it out here and see if the community thinks I'm on the right path. Happy to course-correct based on your feedback.

## Why I'm doing this

1. **I'm using `late` to learn local models.** I've been running it against oMLX to build intuition for how they behave, with the eventual goal of fine-tuning my own. A big part of that learning is *replaying* conversations — including subagents. Today subagent conversations are in-memory only and vanish the moment the subagent finishes. I want every conversation persisted so I can inspect, replay, and eventually use them as training data.

2. **I want parallel research subagents someday.** Imagine spawning 3–5 research subagents in parallel, each with its own isolated history, then consolidating. For that to be safe, each subagent needs a stable, collision-free path *before* any goroutine spawns. The current session store can't guarantee that, so this redesign sets the stage without actually implementing parallel execution yet.

## The latent problems

To make subagent history land safely, two existing issues need fixing first:

| # | Problem | Why it matters |
|---|---------|----------------|
| 1 | **ID race.** Subagent IDs are minted from `len(parent.Children())` — a read under no lock that races the append in `AddChild` under another lock. Two concurrent spawns both see count = 0, both mint `coder-subagent-0`, and the second rename silently overwrites the first. | Count-based IDs collide under concurrent spawns. Parallel subagents are unsafe until this is gone. |
| 2 | **Flat-dir clutter.** One loose `.json` per session in a shared top-level dir. Add subagent history and every subagent drops another indistinguishable file. Grouping, listing, and deleting a session's artifacts all become ad-hoc. | Happens even with sequential subagents. No parent-child relationship on disk. |

Rather than patch each symptom, one layout change takes care of both.

## The high-level design

**One folder per session.** Subagent histories nest as subfolders inside the parent — so `late session list` (top-level walk) never sees them, `late session delete <id>` cascades for free via `rm -r` semantics, and each session folder is owned by exactly one writer.

```
~/.local/share/late/sessions/
├── session-20260803-192905/        # parent — date-based ID unchanged
│   ├── history.json
│   ├── meta.json
│   └── subagents/
│       ├── coder-subagent-0_01J8Q3PT…/   # ULID-suffixed (new)
│       │   └── history.json
│       └── coder-subagent-1_01J8Q3Q0…/
│           └── history.json
└── session-20260803-185121/
    └── …
```

**The ID race is fixed internally** (not a breaking change): subagent IDs are minted under the parent mutex with a ULID suffix via `NextChildID(agentType)`. Two concurrent spawns can't construct the same path. Parent session IDs stay date-based; only subagents gain the ULID suffix — and since subagent history is brand-new, nothing on disk breaks.

**Migration is one-shot and idempotent.** Existing flat `session-*.json` + `session-*.meta.json` files move into a folder named after the ID (the ID itself is preserved — pure layout move). Runs automatically on startup;

## Making the legacy code easy to eventually remove

For backwards compat, the old flat-layout read/migrate code can stay for a while (maybe a few weeks/months?). To keep that cleanup trivial, I'm isolating all of it in a dedicated `internal/session/oldsesh` package and lifting the fallback decision up to `cmd/late`, so `internal/session` itself is already folder-only with zero legacy branches.

The practical payoff: when the deprecation window closes, it's `rm -r internal/session/oldsesh`, drop the two `TODO(deprecate)` glue lines in `cmd/late`, and `go mod tidy`. `internal/session` doesn't change at all. 

## What I'd love feedback on

I'm mostly looking for a gut check — does any of this feel off, or is there a simpler approach I'm overlooking? If you've wrestled with subagent history or persistence before, I'd love to hear what worked, what didn't, and what I probably got wrong.

Beyond all the design questions — I just want to say that `late` is just really fun to run and work with. I'm grateful it exists. Thank you!!!

jpia:
## Summary

Closes #109 

Introduces **opt-in persistence** for subagent conversation histories, restructures sessions into a hierarchical on-disk layout, and fixes a **thread-unsafe subagent-ID generation** bug that would silently cause file collisions under concurrent spawns.

### 1. Hierarchical Storage Layout (`internal/session/paths.go`)

When persistence is enabled, subagent histories are written under the parent session's folder (`<sessions>/<sessionID>/subagents/<childID>.json`). Parent `.json` / `.meta.json` files stay flat and unchanged.

- `SubagentHistoryDir`, `SubagentHistoryPath`, and `RemoveSessionFolder` manage the subagent tree
- `isValidPathElement` rejects `.`, `..`, path separators, and NUL bytes to prevent directory traversal
- `late session delete` also cleans up the subagent folder alongside the parent session

### 2. Thread-Safe ID Generation (`internal/orchestrator/base.go`)

- Added a mutex-protected `NextChildID(typePrefix string)` counter on `BaseOrchestrator`
- `NewSubagentOrchestrator` now mints the child ID **before** creating the session, so the path is known at construction time
- The old lockless `len(parent.Children())` fallback is removed entirely; a subagent **must** have a `*BaseOrchestrator` parent

### 3. Opt-In Persistence (`cmd/late/main.go`, `internal/config/config.go`)

- **CLI flag:** `--save-subagent-histories` (default `false`)
- **Config file field:** `save_subagent_histories` (JSON)
- **Resolution order:** explicit CLI flag > config file > default (`false`)
- There is deliberately **no environment-variable override** — the user must consciously opt in per run or per config

`effectiveSessionID` is derived from the actual history path so resumed sessions reuse their original folder.

### 4. Defensive Hardening

- **Unsafe session IDs:** if `effectiveSessionID` is empty or unsafe, subagent histories fall back to in-memory instead of writing outside the session tree

## Testing

7 test files updated (~1,128 lines) covering:
- subagent orchestrator construction with/without persistence and parent-type enforcement
- path validation (`..`, separators, NUL), resolution, and folder cleanup
- `NextChildID` counter increment and concurrency safety
- config precedence rules and CLI flag wiring
- `deriveEffectiveSessionID` edge cases and session-delete cleanup path

## Backwards Compatibility

- Default behavior is unchanged: without the flag or config field, subagents stay in-memory and legacy flat sessions continue to work
- **No automatic migration** of existing flat files is performed (out of scope)

---

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -255,7 +255,7 @@ late session load <id>     # Resume a previous session
 late session delete <id>   # Delete a session
 ```
 
-By default, subagent conversations are kept in memory and discarded. With `--save-subagent-histories` (or `"save_subagent_histories": true` in the config file), each subagent's history is saved under the session folder — `~/.local/share/late/sessions/<session-id>/subagents/<type>-subagent-<N>.json` — while the parent session files stay in place. `late session delete <id>` removes the subagent folder along with the session.
+By default, subagent conversations are kept in memory and discarded. With `--save-subagent-histories` (or `"save_subagent_histories": true` in the config file), each subagent's history is saved under the session folder — `~/.local/share/late/sessions/<session-id>/subagents/<type>-subagent-<N>.json` — while the parent session files stay in place. `late session delete <id>` removes the subagent folder along with the session. Note that saved transcripts may contain sensitive content (file contents, tool outputs); they are written with user-only permissions.
 
 ## Git Worktrees
 
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -255,8 +255,6 @@ late session load <id>     # Resume a previous session
 late session delete <id>   # Delete a session
 ```
 
-By default, subagent conversations are kept in memory and discarded. With `--save-subagent-histories` (or `"save_subagent_histories": true` in the config file), each subagent's history is saved under the session folder — `~/.local/share/late/sessions/<session-id>/subagents/<type>-subagent-<N>.json` — while the parent session files stay in place. `late session delete <id>` removes the subagent folder along with the session. Note that saved transcripts may contain sensitive content (file contents, tool outputs); they are written with user-only permissions.
-
 ## Git Worktrees
 
 Late is designed for parallel development. You can manage Git worktrees directly to run separate agent instances in isolated environments:
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -242,6 +242,7 @@ You can also create an `.llmignore` file alongside your `.gitignore` to specific
 | `--subagent-max-turns <n>` | Set max turns per subagent (default: 500) |
 | `--append-system-prompt "..."` | Append text to the system prompt (e.g. further instructions) |
 | `--enable-images` | Treat models as supporting images (for none llama.cpp servers) |
+| `--save-subagent-histories` | Persist subagent conversation histories to disk. Off by default (subagent transcripts are large); can also be enabled via `save_subagent_histories` in the config file |
 
 ## Sessions
 
@@ -253,6 +254,8 @@ late session list -v       # Verbose listing with details
 late session load <id>     # Resume a previous session
 late session delete <id>   # Delete a session
 ```
+
+By default, subagent conversations are kept in memory and discarded. With `--save-subagent-histories` (or `"save_subagent_histories": true` in the config file), each subagent's history is saved under the session folder — `~/.local/share/late/sessions/<session-id>/subagents/<type>-subagent-<N>.json` — while the parent session files stay in place. `late session delete <id>` removes the subagent folder along with the session.
 
 ## Git Worktrees
 
--- a/internal/orchestrator/base_test.go
+++ b/internal/orchestrator/base_test.go
@@ -141,9 +141,17 @@ func TestNextChildID_FormatAndMonotonic(t *testing.T) {
 
 	var got []string
 	for i := 0; i < 3; i++ {
-		got = append(got, o.NextChildID("researcher"))
+		id, err := o.NextChildID("researcher")
+		if err != nil {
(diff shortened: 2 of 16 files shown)


internal/orchestrator/base_test.go
```diff
@@ -133,3 +135,74 @@ func TestBaseOrchestrator_ResetStartsNewConversation(t *testing.T) {
...
+	}
+	got = append(got, o.NextChildID("coder"))
+
+	want := []string{
+		"researcher-subagent-0",
+		"researcher-subagent-1",
+		"researcher-subagent-2",
+		// The counter is global per parent, NOT per type.
+		"coder-subagent-3",
+	}
```
jpia, line 162: I wonder if I should change the naming scheme to: 

```
subagent-[index]-[type]
```

docs/quickstart.md
```diff
@@ -254,6 +255,8 @@ late session load <id>     # Resume a previous session
 late session delete <id>   # Delete a session
 ```
 
+By default, subagent conversations are kept in memory and discarded. With `--save-subagent-histories` (or `"save_subagent_histories": true` in the config file), each subagent's history is saved under the session folder — `~/.local/share/late/sessions/<session-id>/subagents/<type>-subagent-<N>.json` — while the parent session files stay in place. `late session delete <id>` removes the subagent folder along with the session. Note that saved transcripts may contain sensitive content (file contents, tool outputs); they are written with user-only permissions.
+
```
jpia: I wonder if this is TMI for the quickstart. I could remove it if it's unnecessary.

internal/orchestrator/base_test.go
mlhher: I think this is fine as it is as subagent counter should be global.

docs/quickstart.md
mlhher: I think a short explanation of the flag is enough inside the table.

PR #131, opened 2026-09-21
feat: subagent control — per-spawn budgets, idle watchdog, per-call tool timeouts

Emasoft:
## Description of Changes

Four features giving the user and the orchestrator explicit control over subagent lifetime and tool call duration:

**Subagent time budgets (default 24h)**
- Every subagent run is bounded by a wall-clock budget: `--subagent-timeout` (default `24h`; `0` = unlimited), also settable persistently as `subagent_timeout` in `config.json`. Precedence is flag > config > default; invalid config values warn and fall back to the 24h default.

**Per-spawn budget on `spawn_subagent`**
- The orchestrator can budget a single run via the new optional `timeout` argument of `spawn_subagent` (e.g. `"45m"`, `"2h"`; `"0"` = unlimited; omitted = the global value). Invalid values return an error result instead of silently changing behavior.

**Idle watchdog**
- Per-subagent activity tracking (stream progress, in-flight tools, nested spawns) drives a watchdog: `--subagent-idle-timeout` (default `15m`; `0` = off) fires a `SubagentIdleEvent` including a transcript probe of the subagent's last activity, surfaced in the TUI. `--subagent-idle-kill-after` (default `0` = notify only) escalates true idleness to a two-stage kill: the in-flight tool is cancelled first, then the agent run.

**Per-call tool timeouts + shell hardening**
- The `bash` tool accepts a per-call `timeout` argument (absent = the 10m global default; `"0"` = unlimited; invalid = error result). Shells run in their own process group (Unix `Setpgid` + group SIGKILL; Windows `taskkill /T /F`) with `WaitDelay=5s`, so grandchild processes that inherit the output pipes can no longer hang `Wait` forever, and cancellation always kills the whole tree. A session kill hook makes in-flight tool calls cancellable mid-wait, which is what lets the idle watchdog (and the user) stop a stuck tool.

## Independence

This PR is independent of #127/#129/#130 — based on main; textually overlaps with #130 (shell hardening, runner budget): merge one, rebase the other.

## Behavior changes
- Subagents now carry a default 24h wall-clock budget (`0` disables); runs exceeding it are cancelled.
- Idle subagents notify the orchestrator/user after 15m of true idleness (stream progress, in-flight tools, and nested spawns all count as activity).
- Idle-kill is opt-in (`--subagent-idle-kill-after`); by default idle subagents are only notified about, never killed.
- `bash` invocations default to a 10m per-call bound; cancellation kills the whole process group.

## Test plan
- [x] `go test ./... -race -count=1` full suite
- [x] `go build ./...`; `GOOS=windows go build ./internal/tool/`; `go vet ./...`; `gofmt -l` clean on all changed files
- [x] New tests: `internal/tool/shell_timeout_test.go` (per-call timeout, `"0"` unlimited, invalid timeout, global timeout expiry and group kill), `internal/tool/subagent_test.go` (per-spawn timeout parsing), `internal/config/config_test.go` (`subagent_timeout` resolution precedence, invalid values, JSON round-trip), `cmd/late/main_test.go` (explicit `--subagent-timeout` flag reporting), `internal/orchestrator/base_idle_test.go` (idle watchdog transitions, idle event, escalation to kill), `internal/executor/inflight_test.go` (in-flight tool kill hook)

## Known limitations
- The `internal/plugin` package has known environmental failures on main (3 HOME/XDG-dependent plugin tests, plus a flaky `TestRunHook_ProcessGroupKillsChildrenOnCancel`) — reproduced on this branch unchanged and unrelated to these features.
- Approval-waiting tool calls count as active (not idle) — a subagent waiting on user approval will not trigger the idle watchdog.
- The global bash timeout (10m) is currently not user-configurable; only the per-call argument overrides it.


mlhher:
I like the thought behind it, seems valuable.

Note that things that directly affect the LLM performance need to be evaluated with care as thus I will likely proceed with #130 first. Also please make sure to note the latest comment in #127, or I can push a commit removing that part for now and then merge. As noted there we can always come back to it.

